PR #28, opened 2015-08-13
Flesh out readme and adapt files to help Contributors build

zrisher:
This PR includes a number of changes to make it easier for outside contributors to work with the project. With these changes, contributors should be able to follow the Readme and get this set up in < 5 minutes.
- Significant rewrite of `README.md` that walks users through initial local configuration and includes a lot of notes from my trials and tribulations getting it running. I've updated this to the latest build, which simplifies the database setup nicely. :smile: 
- Remove `dropq` requirements in `webapp.views` and `webapp.helpers`. There still remain some references in `webapp.tasks` - I haven't actually tackled running the worker on this public version yet, that will be next.
- Add > 0 length string defaults for `MANDRILL_API_KEY` and `SECRET_KEY` so the server will start without them supplied.
- Add an `environment.yml` so `conda env [create|update]` can be run without a file argument.
- Add a version of the requirements file `requirements_nopsycopg.txt` that leaves out the `psycopg2` requirement as well as `django-toolbelt==0.0.1` which also requires psycopg2. I was unsure of the best way to do this - I know the requirements file is used by the production build, so I didn't want to change it. I could instead drop it all in `requirements_dev.txt` and instruct new contributors to simply pip install that one file, but it didn't feel right since not everyone running this for development will only want to use SQLite. IMO any additional requirements that exist only for OSPC's production building and deployment should probably be left out of the main requirements files, but I'm not very close to the process for that so I'll that as something for you guys to ponder.
- Gitignore the `src` folder left behind by Pip when installing django-flatblocks.

@talumbau @MattHJensen Please let me know what I should change/fix!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,127 +1,231 @@
-Taxbrain Web Application
-========================
+# OSPC Webapp
+<!---
+  @todo Add code status indicators?
+    Possible services include codeclimate, travis, codeship, coveralls, inch-ci, hakiri
+    Examples:
+  [![Build Status](https://codeship.com/projects/0e40c6d0-4787-0132-ccf4-025f0fbdfe45/status?branch=master)](https://codeship.com/projects/0e40c6d0-4787-0132-ccf4-025f0fbdfe45)
+  [![Build Status](https://secure.travis-ci.org/angular-ui/bootstrap.svg)](http://travis-ci.org/angular-ui/bootstrap)
+  [![Code Climate](https://codeclimate.com/github/wellbredgrapefruit/asari.png)](https://codeclimate.com/github/wellbredgrapefruit/asari)
+  [![Security](https://hakiri.io/github/wellbredgrapefruit/asari/master.svg)](https://hakiri.io/github/wellbredgrapefruit/asari/master)
+  [![Test Coverage](https://img.shields.io/coveralls/resque/resque/master.svg)](https://coveralls.io/r/resque/resque)
+  [![Inline Doc Coverage](http://inch-ci.org/github/resque/resque.svg)](http://inch-ci.org/github/resque/resque)
+-->
+Webapp is a Python/Django-based web serving application for the Open Source
+  Policy Center. It serves static pages with information on the OSPC and
+  provides GUI access to its apps.
 
-# Deploying To Heroku
+* [Getting Started](#getting-started)
+  * [Requirements](#requirements)
+  * [Installation](#installation)
+  * [Database Setup](#database-setup)
+  * [Debug Flags](#set-debug-flags)
+  * [Bower](#bower)
+* [Running the App](#running-the-app)
+  * [Using Django](#using-django)
+  * [Using Foreman](#using-foreman)
+* [Making Changes](#making-changes)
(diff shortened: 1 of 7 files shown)


webapp/apps/taxbrain/helpers.py
```diff
@@ -69,7 +68,12 @@
     [         1,   '%', 1],  # "%age Tax Decrease",
     [         1,   '%', 1],  # "Share of Overall Change"
 ]
-TAXCALC_RESULTS_BIN_ROW_KEYS = dropq.dropq.bin_row_names
+TAXCALC_RESULTS_BIN_ROW_KEYS = [
+    "less_than_10", "ten_twenty", "twenty_thirty", "thirty_forty",
+    "forty_fifty", "fifty_seventyfive", "seventyfive_hundred",
+    "hundred_twohundred", "twohundred_fivehundred",
+    "fivehundred_thousand", "thousand_up", "all"
+]
```
talumbau: Hmm... is this an intentional change?

zrisher: Yep, I did that to allow the application to start without dropq (which is not listed in the public requirements files). I moved the dropq-related changes to a new PR #51, which this PR is based off now.

zrisher:
Just added another commit that documents how to make CSS & JS changes.

zrisher:
@talumbau I'm working on updating this to the latest master now and have a bunch of questions regarding package dependencies. Happy to go over these with you on Flowdock:

* Can you confirm for me which files are used by Heroku to designate requirements? Is it `.condarc`, `.conda-requirements`, and `requirements.txt`, but not `requirements_dev.txt`? 
* Can you confirm for me which files are used by Travis to designate requirements? Looking at `.travis.yml` it seems to be `.conda-requirements` plus `pytest` and `setuptools`, then `requirements.txt` and `requirements_dev.txt` plus `dropq`, `pytest-django`, and  `pytest-xdist`.
  * Why does Travis's environment (and no other) need us to require `setuptools`?
  * Shouldn't the local contributor's dev environment also include `pytest` and `pytest-django`?
  * Does Travis need to run tests on postgres, or could it use sqlite as is now standard for local developers?
  * Are we still planning on dropping `dropq` as a dependency of this repo?
* Is there any way we could standardize the method of dependency specification for these environments: Heroku (prod), Travis (test), and local (dev)? I.e. maybe a separate `environment .yml` for each environment, which includes all the pip requirements inline under ` - pip:`? Or a single environment file and 3 different pip requirements files? Just spitballing, you would know far better than me the best approach here. I'm just looking for a way for us to specify dependencies in the same manner for each environment, so they can be easily compared and understood. Surely there's a best practice for that with Python.

brendancol:
@MattHJensen @talumbau closing as stale

zrisher:
@brendancol Thanks for taking a look and cleaning this up. After my last comment above I did have a discussion with @talumbau that covered most of the points, but there remained big outstanding questions about `dropq`, and then I stopped working on this repo, so this PR became outdated. Whenever my involvement with this repo ramps back up I will revisit with updated code and understanding.

Issue #3, opened 2015-08-07
Refine sign in flow


ospc-auto:
_From @MattHJensen on August 6, 2015 23:53_

There are a couple of confusing aspects of our sign in process. 
1. The sign up field does not indicate that it is the source of access to TaxBrain
2. Neither the TaxBrain slide on the home page nor on the "apps" page tells you how to sign up. 

Anyone notice other confusing aspects? 

Notes:
- It seems wise that we merged news signup with TaxBrain signup. 
- Later we may want a contributor mailing list/newsletter, but that will be different and won't need to be on the OSPC homepage. 

_Copied from original issue: OpenSourcePolicyCenter/webapp#253_

zrisher:
It seems to me we could fix most of the problems with a few changes to the login page:

There should be a notification at the top if you've been redirected from a resource that requires authentication, i.e. "you must log in to [use taxbrain | view news]"
the login and password fields should be centered and have bottom margin equal to the top
There should be a link to the new account page
There should be some text that explains new accounts are free and we won't track you or send you annoying emails. Probably best accomplished with a flatblock.

MattHJensen:
@amfarrell and @talumbau, do you have any ideas on this one? If there were a quick fix here, I think it would improve the UI quite a bit while we wait for the 2009 puf. (a couple of people have come to me confused about this).

mistakevin:
> There should be a link to the new account page

There is no single link to the account creation page. You have to go through the newsletter subscription first.

The current flow goes:
1. Enter email on homepage (for seemingly **only** the newsletter)
2. Receive confirmation email with new account activation link (mention of authorized use of taxbrain)
3. Click **unique link** in email, directs to account creation page.
4. Create an account. Save. Redirect to a page that suggests returning to home page or taxbrain
  * Also get email with username/password, with mention of taxbrain

We could:
1. Separate the newsletter/account creation
2. Remove the unique "account creation link" email step
3. Replace it with a "Create an Account" link on the homepage
4. After registering, send a "Confirm email address" email with a unique link.
5. Direct user to a "Thank you for confirming" page that mentions taxbrain

zrisher:
Yep, I like all of @mistakevin's suggestions. Newsletters and accounts have different purposes and intended audiences, even if they are linked by User.

MattHJensen:
This makes sense to me as well. We will also want the "create an account" link on the TaxBrain signin page.

talumbau:
[...]

PR #36, opened 2015-09-09
Refine sign in flow

kevincolten:


talumbau:
The two emails seem too similar to me. Text when registering for the page:

```
Welcome! 

Thank you for registering with ospc.org. This is the best way to stay up to date on 
the latest news from the Open Source Policy Center. We also invite you to beta test 
the TaxBrain webapp. 

Username: g10391160 
Password: taxbrain 
```

Text for signing up for mailing list:

```
Welcome! 

Thank you for registering with ospc.org. This is the best way to stay up to date on 
the latest news from the Open Source Policy Center. We also invite you to beta test 
the TaxBrain webapp. 
```

I think for the "sign up" action, the email should say something like "Thanks for signing up for the OSPC mailing list! This is the best way to stay up to date on the latest news from the Open Source Policy Center"

I think the for "register" action, the email should say something like "Thanks for registering! Keep track of the account info below because you'll need it log in to ospc.org. We'll also keep you informed about the latest news from the Open Source Policy Center." (because the "register" action includes getting on the mailing list, it just also gives you an account).

talumbau:
I realized while reviewing this PR and #3 that I don't actually understand the goals we are trying to achieve. I think it's best to disregard what I've written above until I have more clarity

talumbau:
One thing that's true about this PR is that we have 100% removed the confirmation email link, which seems wrong. We definitely want a confirmation email on signing up via email.

mistakevin:
@talumbau This has been updated to match the current specs as I understand them. It is important to note that subscribers and registered users are still in separate tables.

talumbau:
Trying this PR out on latest master. The interactions are simple, which is good, but they have left out important functionality. The following is based on details gleaned from a conversation with @MattHJensen. The new terminology is my own suggestion.

This PR (and the relevant code in current master) are in support of two features: signing up for the newsletter and registering with a username, password, and email address to be a registered user of TaxBrain. We will now use the following terms for these interactions:

 - one *subscribes* to the newsletter
 - one *registers* for the website

Therefore, all buttons that say "SIGN UP" should be changed to say "SUBSCRIBE". "SIGN UP" could indicate either interaction.

1. The *subscribe* flow is:  put in email address -> get email with confirmation link -> click link and go to page that confirms subscribing. This PR has removed the confirmation link -> confirmed subscribe part of the interaction. It should go back in.

2. The *register* flow is: put in username/password/email address -> get email with username/password in it and confirmation link -> click link and go to page that confirms registering. This PR has removed the confirmation link -> confirmed registration part of the interaction. It should go back in.

I think with those fixes this would be good to go.

talumbau:
Actually, let me clarify. The "SIGN UP" button at the top of the site should be changed to "REGISTER" and the "SIGN UP" button next to the "Enter email to subscribe" should be changed to "SUBSCRIBE". That will distinguish to behaviors.

mistakevin:
@talumbau updated buttons

talumbau:
@mistakevin ok great thanks. It looks like the remaining tasks on this PR then are:

- put the confirmation link back in on the *register* interaction

- put the confirmation link back in on the *subscribe* interaction

mistakevin:
@talumbau You mean the confirmation links in the emails? Are they not showing up?

talumbau:
Correct. they are not showing up with this PR.

mistakevin:
I'm getting both of them. Is the spec something else?

screen shot 2015-10-04 at 8 54 07 pm
screen shot 2015-10-04 at 8 55 30 pm

mistakevin:
Now resolves Issue https://github.com/OpenSourcePolicyCenter/webapp-public/issues/11

brendancol:
@talumbau closing as stale

Issue #30, opened 2015-08-15
ENH: the webapp should validate taxcalc and dropq versions upon receiving work


talumbau:
In the response from a worker node, the finished task should indicate versions for `dropq` and `taxcalc` used to compute the values. All work from all workers should match.

talumbau:
closed via #43

Issue #29, opened 2015-08-15
BUG: The webapp should report the taxcalc version from returned results


talumbau:
Now that the dropq workers do all of the work, the version for dropq and taxcalc should be returned in the response, and the results page should display that version. Currently, we get the displayed version from the taxcalc version installed locally. This version may not match the version actually doing the work, and so would lead to a results page where a particular result is displayed with a version number that is inaccurate.

talumbau:
the webapp now imports `taxcalc` and `dropq`, then optionally validates the versions of `taxcalc` and `dropq` from the remote workers. Once we turn strict enforcement on, it will be impossible to get results that come from a different version of `taxcalc` than what is reported.

PR #43, opened 2015-09-12
adds option of enforcing a version check

talumbau:
- If this feature is turned on (currently defaults to off), TaxBrain
  will check if the version of the remote workers is the same as the
  local version of taxcalc, and raise an error if there is not a match


talumbau:
Addresses #29 and #30

PR #49, opened 2015-09-17
Fix 'Edit Parameters' button on results page

talumbau:
- Adds new view 'taxbrain/edit/{id}. We go back to the database
  to get the model parameters and then populate the input screen
  based on that. This avoids any issues with storing input info
  in the session, which we previously had a problem with.


talumbau:
Resolves https://github.com/OpenSourcePolicyCenter/webapp-public/issues/32

PR #51, opened 2015-09-23
remove dependency on dropq to run webserver

zrisher:
These changes are required for a public contributor (who would not have access to `dropq`) to successfully run the web application. Without them, python will throw import errors.

**Public contributors are not able to run the calculator job**
Note that this does not enable a contributor to run the calculator to support submissions of the taxbrain inputs form - I'm not sure what the plan was to tackle that, but it's pretty involved and will likely require some standardization between the output formats from `dropq` and `taxcalc`.

**Statically defined results format**
`TAXCALC_RESULTS_BIN_ROW_KEYS` and `TAXCALC_RESULTS_DEC_ROW_KEYS ` are now statically defined instead of pulled from `dropq`. The definitions are pulled directly from the current version of `dropq`, so should work fine. This would remove our ability to change the output format in `dropq` and have it automatically picked up in `webapp`, except we never actually had that ability: the label variables already statically define their keys. We could make this conditional on the added `dropq_available` variable and if it's available import all the results keys and labels dynamically, but I think the answer to the above issue about running the calculator will probably change how these labels and keys are stored and managed. So for right now this seems appropriate.

### DIFF
--- a/webapp/apps/taxbrain/helpers.py
+++ b/webapp/apps/taxbrain/helpers.py
@@ -1,6 +1,5 @@
 from collections import namedtuple
 import taxcalc
-import dropq
 import os
 import requests
 from requests.exceptions import Timeout, RequestException
@@ -8,6 +7,11 @@ import json
 import pandas as pd
 import time
 
+try:
+    import dropq
+    dropq_available = True
+except ImportError:
+    dropq_available = False
 
 #
 # Prepare user params to send to DropQ/Taxcalc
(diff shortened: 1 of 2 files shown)


talumbau:
@zrisher, I'm going to let this hang out here for a while, because the intention is to open-source `dropq` very soon. That should resolve the problem of letting contributors run w/o issue. If we have to push back the dates on making `dropq` open source, this seems like a reasonable way to go.

zrisher:
@talumbau Makes sense, thank you!

PR #55, opened 2015-09-29
Taxsaveinputs form validation 2

zrisher:
This includes https://github.com/OpenSourcePolicyCenter/webapp-public/pull/53

This PR allows the webapp to enforce the parameter validation specified in `params.json` on form POST requests, per https://github.com/OpenSourcePolicyCenter/webapp-public/issues/52.

I tried it out with the current validations in params.json and it seems to work great, but I'd love some feedback from others, especially on the error messages.


webapp/apps/taxbrain/helpers.py
```diff
@@ -1,14 +1,52 @@
 from collections import namedtuple
 import taxcalc
-import dropq
```
talumbau: Hmm.. when I tried this out, I needed to restore this line so that I could use dropq later in the file.

talumbau:
This looks good to me, other than the issue about importing `dropq` in `helpers.py`. When I added that line back in, it was good to go and the validation worked as expected. Thanks @zrisher

talumbau:
My plan would be to merge this soon, so if you can add another commit to re-add the import of dropq that's great. otherwise, I will submit a follow-on PR that has the additional commit.

zrisher:
Thanks for taking care of that, yes my apologies I think that line belonged to my dropq commit and snuck into this one.

PR #63, opened 2015-10-13
Remove permission

talumbau:
Allows a user to use TaxBrain without being logged in.

Issue #59, opened 2015-10-07
new subsection under itemized deductions. 


MattHJensen:
We should add a new "Surtax on Itemized Deduction Benefits above an AGI Threshold" subsection in the itemized deductions section on the TaxBrain input page that takes advantage of the new features that @amy-xu added in https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/361. 

This should be slipped under the itemized deduction limitation subsection like this:

![image](https://cloud.githubusercontent.com/assets/8301092/10345754/c3e33c88-6cf9-11e5-95da-7eecf5b9d6ca.png)

A few new things here: 
- There is a new UI element, the checkboxes for which itemized deductions are subject to the limitation.
- The checkboxes are all coming from one parameter, _ID_BenefitSurtax_Switch.
- There should be a flatblock to accommodate a description for a subsection (currently we only have these between sections.)

@mistakevin, let me know if there is any other information that would be helpful. 

This feature allows users to implement something close to the proposal outlined in this paper: http://www.nber.org/papers/w16921.pdf. The idea is that if we calculate a taxpayer's taxes allowing him to take his itemized deductions, and calculate his taxes again without allowing him to take take itemized deductions, we can identify the benefit he derived from those itemized deductions as the difference in the two tax liabilities. Feenberg, Feldstein and MacGuieneas propose limiting that tax benefit to some fraction of adjusted gross income --they propose 2% (actually, they propose more than just itemized deductions, but we are sticking with itemized deductions). 

Another way to think about it, and the way that we will be describing it to our users, is that there is an additional tax, a surtax, on the benefit the taxpayer gets from their itemized deductions. In the FFM proposal, the surtax rate is 100%. However, they allow taxpayers to tax a credit against that 100% tax, where the credit amount is 2% of adjusted gross income. Then you can think of current law for this surtax as either having a 100% credit rate or a 0% tax rate (we use the 100% credit rate approach). 

cc @talumbau @amy-xu @feenberg

mistakevin:
@MattHJensen How can setup my taxcalc to use the latest master from github?

MattHJensen:
@mistakevin I believe you need to use the latest data file. I just sent you an email with a link.

PR #66, opened 2015-10-15
Kevin additions

talumbau:
Quite a bit here:

- #59 from @mistakevin 
- Enable some Behavior Effects
- Adds Growth Assumptions input features.
- also includes commits for #63


talumbau:
added additional commits to allow Edit Parameters to properly populate switches/radio buttons.

Issue #62, opened 2015-10-12
output tables take advantage of payroll taxes


MattHJensen:
We can now calculate payroll taxes. For our tables to take advantage of this we should make some modifications:

Revenue table:
- [x] header -> Total Liabilities Change by Calendar Year
- [x] the total revenue row should be relabelled "Individual Income Tax Liability Change"
- [x] a new row should be labelled "Payroll Tax Liability Change" 
- [x] a new row should be labelled "Combined Payroll and Individual Income Tax Liability Change" 

Diagnostic tables: 
- [x] the "Revenue" column should be relabelled "Individual Income Tax Liabilities"
- [x] a new column should be labelled "Payroll Tax Liabilities" 
- [x] a new column should be labelled "Combined Payroll and Individual Income Tax Liabilities

Difference tables: 
- [x] "Individual Income Tax:" should be prepended to the existing table names, so an example name would be, "INDIVIDUAL INCOME TAX: DIFFERENCE BETWEEN BASE AND USER PLANS BY AGI DECILE"
- [ ] A new set of tables should be added for "Payroll Tax:.."
- [ ] A new set of tables should be added for "Combined Payroll and Individual Income Tax:..."

talumbau:
> a new row should be labelled "Payroll Tax Liability Change"
>  a new row should be labelled "Combined Payroll and Individual Income Tax Liability Change"

In order to add these rows, I need to know the calculation to do. For the current single row in this table, we do the calculation here:

https://github.com/OpenSourcePolicyCenter/dropQ/blob/master/dropq/dropq.py#L122

which looks like this:
```
dec_sum = (df2['tax_diff_dec']*df2['s006']).sum()
```

so just the difference in tax between the two plans for each record weighted by `s006` and then summed. What is the correct calculation for the "Payroll Tax Liability Change" row and the "Combined Payroll and Individual Income Tax Liability Change"? cc @Amy-Xu

Amy-Xu:
@talumbau Yep I saw that line. The other two rows would be calculated in the same way but replacing `_ospctax` with `_payroll` and `_combined` respectively. It's not difficult to add those calculations in `dropq.py`, but I was hesitant because I realized in the `helper.py` this table's if-loops need to be replaced by a for loop. I don't think I want to do that without being able to test it on my computer. I was wondering do you mind make changes to `helper.py` on your end? I can just add the calculations of the two rows here in dropq.py

talumbau:
[...]

PR #67, opened 2015-10-15
Changed table headers and added payroll column

Amy-Xu:
Checked items in #62 are included in this PR. Everything is in line with [this PR](https://github.com/OpenSourcePolicyCenter/Tax-Calculator/pull/392) to taxcalc.

### DIFF
--- a/webapp/apps/taxbrain/helpers.py
+++ b/webapp/apps/taxbrain/helpers.py
@@ -165,7 +165,7 @@ TAXCALC_RESULTS_TABLE_LABELS = {
     'df_dec': 'Difference between Base and User plans by expanded income decile',
     'mX_bin': 'Base plan tax vars, weighted avg per expanded income bin',
     'mY_bin': 'User plan tax vars, weighted avg per expanded income bin',
-    'df_bin': 'Individual Income Tax: Difference between Base and User plans by AGI decile',
+    'df_bin': 'Individual Income Tax: Difference between Base and User plans by expanded income bin',
     'fiscal_tots': 'Total Liabilities Change by Calendar Year',
 }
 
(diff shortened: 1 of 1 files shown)


webapp/apps/taxbrain/helpers.py
```diff
@@ -162,8 +165,8 @@ def string_to_float_array(s):
     'df_dec': 'Difference between Base and User plans by expanded income decile',
     'mX_bin': 'Base plan tax vars, weighted avg per expanded income bin',
     'mY_bin': 'User plan tax vars, weighted avg per expanded income bin',
-    'df_bin': 'Difference between Base and User plans by expanded income bin',
-    'fiscal_tots': 'Total Revenue Change by Calendar Year',
+    'df_bin': 'Individual Income Tax: Difference between Base and User plans by AGI decile',
```
talumbau: Hmm.. something has gone wrong here. Any name that ends with `_bin` refers to data that has been grouped by an income bin. Here, you are saying that `df_bin` is referring to data that is grouped by AGI_decile. If I understand correctly, the line should be:
```
'df_bin': 'Individual Income Tax: Difference between Base and User plans by expanded income bin',
```

@Amy-Xu do you see what I mean?

Amy-Xu: Yea, haven't been able to double check with @MattHJensen but I think you're right. Will change this in a sec.

MattHJensen: Yes, should be expanded income bin for now. That said, I hope we can generalize this later on so the user can decide whether to use _expanded_income or c00100 as the tab variable for tables. https://github.com/OpenSourcePolicyCenter/webapp-public/issues/35#issuecomment-148811619

talumbau:
this looks good. Merging.

PR #81, opened 2015-10-20
Payroll taxes section

talumbau:

PR #82, opened 2015-10-22
Cpi flag fix

### LINKED ISSUE #69: Social security taxability CPI flags are broken
The social security taxability thresholds are not cpi inflated under current law, which is properly accounted for in current_law_policy.json, but the flags are shown as True on the webapp input page. 

image

HT @viard.

talumbau:
Depends on #81 

Fixes #69


talumbau:
This PR fixes the CPI flag defaults behavior on the input page. Previously, every single CPI flag defaulted to 'on' on the input page. This makes it so that if the `cpi_inflated` value is False in the default JSON file in taxcalc, the flag will be off here. Also, if the non-default value is chosen, that non-default value is correctly propagated through when the job is submitted. The only thing that is lost in this PR is that in the current site, when the non-default is selected, a small piece of text appears that displays the default value. Because the current site improperly considers every flag as default (at least in the UI) this nice feature is actually not helpful because it is giving wrong information. I would much rather merge this PR, which now shows visually the proper defaults, and then ask @zrisher or @mistakevin to take a look at restoring the nice behavior of 'clicking to change the default displays a bit of text that says "Default: <default value>"'

MattHJensen:
@talumbau, I'm not sure if this is related or if we should open a new issue, but a user said: 

> 3) Tried to increase child credit from $1000 to $2000.  If I enter just "2000" there is only one year of revenue change. With the intent of seeing revenue effect of a permanent change I entered 10 values separated by commas. This seemed to work (that is, "2000,2000,2000 . . ."). Is that the correct way to do multi-year parameter changes?

He's right that the proper behavior of propagating the param change into future years is not working. See http://www.ospc.org/taxbrain/527/

talumbau:
I just confirmed the behavior you are describing on www.ospc.org, but I can't reproduce the behavior on my machine just yet. I'll take a deeper look.

MattHJensen:
Interestingly, this propagation problem does not appear to affect *all* parameters. Personal exemption changes do work: http://www.ospc.org/taxbrain/540/

talumbau:
@MattHJensen I have looked into this and discovered that what is happening is designed behavior. We (or really Matt and perhaps others who use TaxBrain) will have to decide if they really want this behavior. Here is what is going on:

The idea for the inputs of TaxBrain (and, actually, specifying a reform calculator) is that you overwrite existing policy, as defined in current_law_policy.json. For `_II_em` ("Personal and dependent exemption amount") we have a single value that is specified for 2013. Its `cpi_inflated` flag is set to True, so if nothing else is specified, we get the following year values by inflating with the pre-defined inflation rates. If a reform specifies a new value (in our case, typically starting in 2015) that new value will be used for the calculation of the following years.

For a parameter like `_CTC_c`, "Maximum child tax credit per child", the current policy has values for up to 2018. So, when a user specifies one year's reform values, we do not by default overwrite the following years, since a value is encoded in the current policy. In order to get the "broadcasting" of the user's choice to the rest of the budget window, the user's choice must go at least as far as what is specified in the current policy. So, in order to get the behavior of a 2000 dollar credit for the 10 years, the user would have to specify the 2000 credit for 2018, because we already have a value up until that point. For TaxBrain, you have to specify a sequence of values, so the user would have to specify 2015, 2016, 2017, and 2018. Does that make sense?

I'm not saying this is the best way to do it, just that the behavior is what we would expect if someone ran the calculator and specified a reform dictionary of `{2015: {'_CTC_c':[2000.0]} }`.

MattHJensen:
@talumbau, I understand what you are saying. The CTC max is more complicated because it is scheduled to decrease to 500 in 2018 under current law.  

I think the proper behavior for tax-calculator should be that if a user overwrites any year's value, even if it isn't the last year's, the user's value should be broadcast to the rest (w or wo CPI inflation), overwriting anything that is currently there. So changing the default value in webapp from 1000 to 2000 would replace all future values, including 2018's 500, with 2000. This would be step one. 

The second step would be to prefill the input page with all values up to and including the stable one. So the displayed default for child tax credit max when when the start_year is 2015 would be 1000, 1000, 1000, 500. The displayed default when the start_year is 2013 would be 1000, 1000, 1000, 1000, 1000, 500.

Does this seem intuitive to you? 

In between step 1 and step 2 there will be a confusing issue of, how do you get 1000s for all years, since the webapp won't recognize an entry of 1000 as a reform and so it will keep the 1000, 1000, 1000, 500, but that minor confusion would get resolved w step 2. But getting step 1 finished incorporated even if step 2 isn't yet still seems like an improvement. 

cc @Amy-Xu @feenberg

talumbau:
Does this change in behavior apply to just Tax Brain or both Tax Brain and taxcalc?

MattHJensen:
The 'Step 1' change I believe should be adopted in tax-calculator (and also, therefore, webapp). The 'step 2' change I believe is webapp specific.

MattHJensen:
cc @martinholmer on the conversation starting from https://github.com/OpenSourcePolicyCenter/webapp-public/pull/82#issuecomment-150628189

MattHJensen:
At @martinholmer's suggestion, I added a 2015 reform of _CTC_c to `test_multi_year_reform()` in `test_policy.py`, changing the value from 1000 to 2000. I also added a `check_ctc_c` function, which I call in `test_multi_year_reform()`. From this testing it seems like the Policy class `implement_reform` is already handling things how I describe that it "should" in step 1 here: https://github.com/OpenSourcePolicyCenter/webapp-public/pull/82#issuecomment-150635128. 

Is that what you are finding @talumbau? 

```
def test_multi_year_reform():
    """
    Test multi-year reform involving 1D and 2D parameters.
    """
    # specify dimensions of policy Policy object
    syr = 2013
    nyrs = 10

    # specify assumed inflation rates
    irates = {2013: 0.02, 2014: 0.02, 2015: 0.02, 2016: 0.03, 2017: 0.03,
              2018: 0.04, 2019: 0.04, 2020: 0.04, 2021: 0.04, 2022: 0.04}
    ifactor = {}
    for i in range(0, nyrs):
        ifactor[syr + i] = 1.0 + irates[syr + i]
    iratelist = [irates[syr + i] for i in range(0, nyrs)]

    # instantiate policy Policy object
    ppo = Policy(start_year=syr, num_years=nyrs, inflation_rates=irates)

    # confirm that parameters have current-law values
    assert_array_equal(getattr(ppo, '_AMT_thd_MarriedS'),
                       Policy.expand_array(np.array([40400, 41050]),
                                           inflate=True,
                                           inflation_rates=iratelist,
                                           num_years=nyrs))
    assert_array_equal(getattr(ppo, '_EITC_c'),
                       Policy.expand_array(np.array([[487, 3250, 5372, 6044],
                                           [496, 3305, 5460, 6143],
                                           [503, 3359, 5548, 6242]]),
                                           inflate=True,
                                           inflation_rates=iratelist,
                                           num_years=nyrs))
    assert_array_equal(getattr(ppo, '_II_em'),
                       Policy.expand_array(np.array([3900, 3950, 4000]),
                                           inflate=True,
                                           inflation_rates=iratelist,
                                           num_years=nyrs))
    assert_array_equal(getattr(ppo, '_SS_Earnings_c'),
                       Policy.expand_array(np.array([113700, 117000, 118500]),
                                           inflate=True,
                                           inflation_rates=iratelist,
                                           num_years=nyrs))
    assert_array_equal(getattr(ppo, '_CTC_c'),
                       Policy.expand_array(np.array([1000, 1000, 1000,
                                                     1000, 1000, 500]),
                                           inflate=False,
                                           inflation_rates=iratelist,
                                           num_years=nyrs))

    # specify multi-year reform using a dictionary of year_provisions dicts
    reform = {
        2015: {
            '_AMT_thd_MarriedS': [60000],
            '_CTC_c': [2000]
        },
        2016: {
            '_EITC_c': [[900, 5000, 8000, 9000]],
            '_II_em': [7000],
            '_SS_Earnings_c': [300000]
        },
        2017: {
            '_AMT_thd_MarriedS': [80000],
            '_SS_Earnings_c': [500000], '_SS_Earnings_c_cpi': False
        },
        2019: {
            '_EITC_c': [[1200, 7000, 10000, 12000]],
            '_II_em': [9000],
            '_SS_Earnings_c': [700000], '_SS_Earnings_c_cpi': True
        }
    }

    # implement multi-year reform
    ppo.implement_reform(reform)
    assert ppo.current_year == syr

    # move policy Policy object forward in time so current_year is syr+2
    #   Note: this would be typical usage because the first budget year
    #         is greater than Policy start_year.
    ppo.set_year(ppo.start_year + 2)
    assert ppo.current_year == syr + 2

    # confirm that actual parameters have expected post-reform values
    check_amt_thd_marrieds(ppo, reform, ifactor)
    check_eitc_c(ppo, reform, ifactor)
    check_ii_em(ppo, reform, ifactor)
    check_ss_earnings_c(ppo, reform, ifactor)
    check_ctc_c(ppo, reform)

    # end of test_multi_year_reform with the check_* functions below:

def check_ctc_c(ppo, reform):
    """
    Compare actual and expected _CTC_c parameter values
    generated by the test_multi_year_reform() function above.
    """
    actual = {}
    arr = getattr(ppo, '_CTC_c')
    for i in range(0, ppo.num_years):
        actual[ppo.start_year + i] = arr[i]
    assert actual[2013] == 1000
    assert actual[2014] == 1000
    e2015 = reform[2015]['_CTC_c'][0]
    assert actual[2015] == e2015
    e2016 = e2015
    assert actual[2016] == e2016
    e2017 = e2016
    assert actual[2017] == e2017
    e2018 = e2017
    assert actual[2018] == e2018
    e2019 = e2018
    assert actual[2019] == e2018
```

talumbau:
yes Matt you are right. The reason we see the unexpected behavior is because of some pre-processing on the input that happens before the calculation is done. Essentially, we combine the user's input with the default parameters and the end result is the behavior I describe. This was done to accommodate the case of a user only entering a subset of some multi-part parameter and other odd cases, but it clearly has this undesirable behavior, since in the case of `_CTC_c`, specifying just the 2015 behavior is the same as specifying a reform dictionary of `{2015: {'_CTC_c':[2000.0, 1000, 1000, 500]} }`, which is bad, whereas the user would likely assume that just specifying the 2015 behavior in TaxBrain is the same as specifying a reform dictionary of  `{2015: {'_CTC_c':[2000.0]} }`. I will make the necessary changes

webapp/apps/taxbrain/tests.py
```diff
@@ -29,6 +29,18 @@ def test_convert(self):
         assert ans['_II_brk2'] == exp
         assert len(ans) == 1
 
+    def test_package_up_vars_CTC(self):
+        values = {"CTC_c": [2000.0]}
+        ans = package_up_vars(values)
+        exp =  [2000.0]
+        assert ans['_CTC_c'] == exp
```
talumbau: @MattHJensen here is the updated behavior: user entry of "2000" for Child Tax Credit gives back a JSON dictionary like `values` which is converted to `ans` in the above. The actual reform dictionary will then be `{2015: {'_CTC_c': [2000.0]}}`, which will then 'broadcast' as expected, which is what Martin's test shows.

MattHJensen:
@talumbau, do you think there should be an issue for this:

>I would much rather merge this PR, which now shows visually the proper defaults, and then ask @zrisher or @mistakevin to take a look at restoring the nice behavior of 'clicking to change the default displays a bit of text that says "Default: <actual default value>"'

talumbau:
Agreed. Created #88

PR #89, opened 2015-10-27
Remove job IDs from session - store on model

talumbau:
- Job IDs for processing jobs are moved from the session to being stored
  on the model. This allows the user to run multiple jobs in different
  tabs in the browser. W/o this, the different jobs in different browser
  tabs would conflict and all tabs would get the same completed job.

  This fix will cause more queries to the database, but it's the most
  correct way to handle processing jobs


talumbau:
Fixes https://github.com/OpenSourcePolicyCenter/webapp-public/issues/87

PR #91, opened 2015-10-27
celery status server with flask

PeterDSteinberg:
Defines the following urls at port 5070:

/register_job   
* ticket_id
* email
* callback, a url that startswith http:// or http:// will be added

/celery_status_server
* ticket_id or no args

/pop_ticket_id
* ticket_id, call this after you are done tracking a ticket_id and have gotten its callback response.

### DIFF
--- a/celery_status_server.py
+++ b/celery_status_server.py
@@ -1,0 +1,126 @@
+#celery_status_server.py
+
+from flask import Flask, request, make_response
+from celery import Celery
+from celery.app.control import Inspect
+from celery.result import AsyncResult
+import datetime
+import requests
+import os
+import json
+import time
+import sys
+import subprocess as sp
+from functools import partial
+import example_async
+"""
+nohup redis-server &
+nohup celery -A example_async worker -P eventlet -l info &
+"""
+server_app = Flask('celery_status_server')
+port = 5070
+server_url = "http://localhost:{0}".format(port)
+
+TICKET_IDS_WATCHED = {}
+m = '--module'
+if not m in sys.argv:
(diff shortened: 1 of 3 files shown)


talumbau:
@PeterDSteinberg I realized we may be thinking about this server in different ways. That is, we may be thinking about which pieces of functionality go where. My thoughts are in this diagram:

imag0160

Here, the new `notifier` server runs next to the Django process on Heroku like we talked about. It has a `SUBMIT` endpoint, which is your `register_job`, so that part is the same. For checking on the job, I think it could reuse the `dropq_query_results` endpoint on the flask server on the worker nodes:

https://github.com/OpenSourcePolicyCenter/dropQ/blob/master/flask_server.py#L43

so job checking is a service provided by the AWS workers, not something that this `notifier_server` does. It still needs the part of 'once a minute, goes ask if the jobs you are checking on are done. So this makes me thing we actually do need a thread, because the `register_job` endpoint will be hit by the django process but then that data will be needed to check every minute if the job is done. So one way to do that is to spawn a thread that only wakes up every minute to check with the worker nodes if the work is done. After that, it would hit some kind of /MAKE_DYNAMIC endpoint on the django process, which would populate the database with what is needed for displaying the results of the dynamic run. Then, it would send an email to the registered user. The user would get an email with a link so that when they click it, then go to the /DYNAMIC endpoint I show in the diagram.

So, these are my thoughts. I don't know if this is the best way, but I wanted to make sure you at least understood what I was saying. Let me know what you think!

talumbau:
this work is now over in dropq.

Issue #71, opened 2015-10-17
on homepage, "from the lab is off centered" 


MattHJensen:
![image](https://cloud.githubusercontent.com/assets/8301092/10560142/e973b99a-74cf-11e5-9857-11068b679912.png)

cc @talumbau, @mistakevin, @zrisher

mistakevin:
I tried to come up with a quick style fix, but it turned out to be a css rabbithole. I think this page should be laid out using the bootstrap grid, especially since development will look different than production due to flatblocks, we need to be as responsive as possible.

zrisher:
I agree with @mistakevin, I think the positioning on this page needs some refactoring. Simply adding `text-align: center` to this element's style (or better yet the bootstrap class of `text-center` to the element) has no effect because of the relative positioning being applied to the surrounding elements to accomplish the background that stops halfway through the TaxBrain pane. If we forgo that background trick, fixing this is as simple as 

* remove `margin-bottom: 200px;` from the container `section.home-display`
* remove `bottom: -200px;` and `margin-top: -200px;` from `div.home-display-app` and add `margin-bottom: 50px`

![image](https://cloud.githubusercontent.com/assets/1958130/10803397/30f9e2be-7d7f-11e5-8323-d102949c1d94.png)

There's a lot of relative positioning going on in this page (and in the Taxbrain section I believe) because all of these pages had static content when we received the styles from the initial designers. Now that we've moved to a dynamic content model, more fully adopting bootstrap would make a lot of sense.

MattHJensen:
>more fully adopting bootstrap would make a lot of sense. 

How big of a project is that and how would you recommend prioritizing it with ongoing work? 

>If we forgo that background trick, fixing this is as simple

In the meantime, I'd forgo aesthetics for the bug fix.

zrisher:
> In the meantime, I'd forgo aesthetics for the bug fix.

See above, submitted a hotfix.

> How big of a project is that and how would you recommend prioritizing it with ongoing work?

Fairly large project as it involves rewriting a good chunk of the CSS, which brings with it the need to test on a lot of browsers and resolution sets. I would include it in a pass that fixes the dependency on `staticfiles` and removes the unused CSS we still have in the app. I would prioritize it after anything that directly affects the usability or functionality of the calculator pages.

talumbau:
@johneo please read over this issue and let me know your thoughts. thanks!

johneo:
[...]

PR #93, opened 2015-10-28
Hotfix "from the lab" offcenter on HP - fixes #71

zrisher:
Compiling from less this time used 3-character representations of repeated hex colors instead of 6-character, which makes up the majority of changes in this PR. I'm hoping this is just because I installed the latest version of lessc, and these changes are good to incorporate. @mistakevin, @talumbau Let me know if you agree.

See https://github.com/OpenSourcePolicyCenter/webapp-public/commit/b438a927e166ce1ba29337297e4af7206f68d8f6 for the actual effective changes that fix #71.

Note I tested this locally, but hard to say if it's actually fixed due to #86.

PR #100, opened 2015-11-10
Issue 71: Centering the heading "FROM THE LAB"

johneo:
Addresses issue #71. See the issue for the screenshot.

### DIFF
--- a/static/css/site.css
+++ b/static/css/site.css
@@ -8129,6 +8129,11 @@ fieldset[disabled] .form-group.cpi-group .btn.btn-cpi {
   text-transform: uppercase;
   color: white;
 }
+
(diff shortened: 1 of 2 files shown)


johneo:
@talumbau should this be put on ice?

talumbau:
No, or at least, I'm not sure. It turns out we also have https://github.com/OpenSourcePolicyCenter/webapp-public/pull/93 which is another fix for the issue. I'm not sure which is the better one to choose. What do you think?

johneo:
#93 looks good, it centers the heading without adding the extra gap that this one does. Also, it seems to do some code cleanup.

talumbau:
closing in favor of #93

PR #104, opened 2015-11-18
Added social sharing button feature.

johneo:
https://github.com/OpenSourcePolicyCenter/webapp-public/issues/48


talumbau:
@johneo this looks cool! @MattHJensen, here is a screenshot of using the tool:

screen shot 2015-11-23 at 4 21 38 pm

Do you have a suggestion for the default text for a tweet/FB post/etc.? Or is this default text OK?

talumbau:
Oh, it would of course give the URL as www.ospc.org.

talumbau:
@johneo the Google+ button appears to create a blank G+ post. The email button doesn't appear to do anything. This is with Chrome on Mac OS X. Do these work for you?

johneo:
@talumbau will take a look

johneo:
@talumbau the share button library does not support the pre-filled comment for G+ because Google's URL for doing it only accepts the the link as a query parameter. It can be done by adding Google's JS library for G+, though the button has no direct support for it. Should I look into that? Also, the share URL is set to http://www.ospc.org/taxbrain when the site runs as localhost because G+ refuses localhost address. In an environment with a normal domain, it will share window.location.href.

As for the mail link not working, my searches seem to indicate that it has to do with the end user's browser setting, though it is not conclusive. Opening the URL in a new window will trigger my mail client to open a new message window with correctly filled-out subject and body. I'll look further into it.

talumbau:
Ok, thanks for the info. Can we remove the "mail" share button? If yes, then I think it is functional enough to use as-is.

johneo:
@talumbau Done.

PR #110, opened 2015-11-25
Fixed #26: PDF being generated from the login screen instead of tax results.

### LINKED ISSUE #26: PDF Breaks With Login Requirement
_From @theofanislekkas on February 18, 2015 15:56_

The pdf functionality is not producing the desired result while the login decorators are in place.  The pdf is returning an image of a page with the login fields instead of the table & the params used.

### NOTE: The pdf functionality is still not live

_Copied from original issue: OpenSourcePolicyCenter/webapp#91_

johneo:
Resolves #26 @talumbau

### DIFF
--- a/templates/taxbrain/results.html
+++ b/templates/taxbrain/results.html
@@ -72,7 +72,7 @@
               Save As <span class="caret"></span>
             </button>
             <ul class="dropdown-menu" role="menu">
-              <li><a href="/taxbrain/pdf">PDF Report</a></li>
+              <li><a href="/taxbrain/{{ unique_url.pk }}/pdf/">PDF Report</a></li>
               <li><a href="/taxbrain/{{ unique_url.pk }}/output.csv/">Outputs as CSV</a></li>
               <li><a href="/taxbrain/{{ unique_url.pk }}/input.csv/">Inputs as CSV</a></li>
             </ul>
--- a/webapp/apps/taxbrain/urls.py
+++ b/webapp/apps/taxbrain/urls.py
@@ -5,10 +5,10 @@ from .views import personal_results, tax_results, output_detail, csv_input, csv_
 
 urlpatterns = patterns('',
     url(r'^$', personal_results, name='tax_form'),
+    url(r'^(?P<pk>\d+)/pdf/$', pdf_view),
     url(r'^(?P<pk>\d+)/output.csv/$', csv_output, name='csv_output'),
     url(r'^(?P<pk>\d+)/input.csv/$', csv_input, name='csv_input'),
     url(r'^(?P<pk>\d+)/', output_detail, name='output_detail'),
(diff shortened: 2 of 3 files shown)


talumbau:
I don't understand the comment with this commit. What does it mean to say that the "PDF [is] generated from the login screen instead of tax results"?

johneo:
That's the issue that the PR fixes. The pdf generation requires user login. The view code was using the referrer URL to convert to PDF, which would be the login screen instead of the results screen, if the user is not logged in. The PR uses the results Id to look up the URL instead of the referrer URL so the login prompt does not affect the pdf output.

talumbau:
Oh I understand, thanks! I will test this locally and see how it goes.

talumbau:
OK, this works great for me locally, although the `pdf_view` still has the `permission_required` decorator, which I would prefer to remove. I made some commits to make this work on Heroku, then I deployed it to a test application. Here is what I found: the call to `pdfkit.from_url` (i.e. the real work) takes longer than 30 seconds on the under-powered VM that hosts the actual website. Prior to this, its only job has been to display the pages and parse the data from the worker nodes to make the output tables. It's apparently a much bigger job to create the PDF and return it. Heroku is set up to give an application error if any response takes longer than 30 seconds. So, the options to make this work are:

- try a more powerful (non-free) Heroku VM for hosting the site itself) and see if any of those work
- change the work pattern to be similar to the current dropq calculation: we move the PDF view over to a worker node, then poll every five seconds to see if the work is done, then pull the data down and return it.

I would have to do some experiments to see if the first option works. If it does, there would be essentially no follow on work (just merging this PR for the most part) but there would be a small increase in monthly cost of the site.

The other option would require more work to refactor where the PDF is created.

This is not the highest priority this week so I'm tempted to just leave this for now. @MattHJensen, do you have thoughts on this, specifically which option you would prefer that I explore?

-T.J.

MattHJensen:
> try a more powerful (non-free) Heroku VM for hosting the site itself) and see if any of those work 

This sounds like a good short/mid term solution.

talumbau:
Unfortunately, it seems that the issue is a hang on Heroku, not an underpowered VM. I have investigated pretty fully and can't figure out why the subprocess call to `wkhtmltopdf` hangs on Heroku (subprocess call happens inside Python package `pdfkit`. The options are to either open a ticket on Heroku or switch to a different technique to generate the PDF. I think switching to a plug-in on Heroku for PDF generation is probably OK because if we switch away from heroku at some point, then we can just use `wkhtmltopdf` anyway.

johneo:
@talumbau I'm sure you've already looked at this: https://github.com/pdfkit/pdfkit/wiki/Use-with-Heroku

Is this currently how we have pdfkit deployed on Heroku?

talumbau:
no longer relevant. closing.

PR #113, opened 2015-11-30
48 social 2

talumbau:
based on #104, updated with:

- merge of master
- changed text when sharing a tweet
- copied assets changed in static/ to staticfiles/

I'm not sure if the last one is necessary. @mistakevin @zrisher can you clarify whether it is necessary to keep the files in staticfiles/ consistent with static/ when making a change to the css and/or js? Thanks!


staticfiles/js/site.js
```diff
@@ -4,3 +4,29 @@
 $(function () {
   $('[data-toggle="tooltip"]').tooltip();
 });
+
+$(document).ready(function(){
+	var urlToShare = window.location.href.indexOf('http://localhost:8000') > -1 ? 'http://www.ospc.org/taxbrain' :window.location.href;
+	new ShareButton({
+		title: 'TaxBrain Results',
+		description: 'Check out my output from OSPC TaxBrain, a tax policy simulator: ' + urlToShare,
```
MattHJensen: How about, 'Check out my output from TaxBrain, an open-source tax policy simulator.'

mistakevin:
You should make changes in `staticfiles/` locally, then run `python manage.py collectstatic` to copy them into `static/`, then commit both.

johneo:
Is there a need for committing the collected static files into the repo, along with the source static files? The Django docs suggest running `collectstatic` at deployment time, preferably automated: https://docs.djangoproject.com/en/1.8/howto/static-files/deployment/

Heroku's doc: https://devcenter.heroku.com/articles/django-assets#automatic-collectstatic

Also, the target directory name could use some improvement. `/static` and `/staticfiles` are too similar, especially if they both get committed to the repo. Throwing in a word like "deploy" or "prod" could clarify things a bit?

talumbau:
tested this on Heroku test app. Looks good. Merging. Thanks @johneo!

PR #116, opened 2015-12-19
Ogusa fixes2

talumbau:


talumbau:
Additional fixes for:
 - deployment of ogusa on Heroku
 - multiple ogusa worker nodes

PR #119, opened 2016-01-04
Zrisher fix hp

talumbau:
Updated version of @zrisher fix to 'FROM THE LAB' off-center on main page.


talumbau:
Oops. There's no need to do this as there's no conflict to resolve with what's in master. closing.

PR #123, opened 2016-01-07
118 errors as warnings

### LINKED ISSUE #118: A user should be able to bypass TaxBrain parameter validation
The idea is that sometimes one wants to force a parameter setting that is outside the allowed parameter ranges specified in the validation metadata. It is still desirable to give an error message when a parameter is set to something outside of the allowed range, but there should be a way to circumvent this setting.

A simple idea that has relatively low effort would be that on the input page, the first time a user sets a value outside of the parameter range, the error page shows up just as it does now. The change would be that on the error page, a user could enter any values they wish, which would not be validated. So, if the user entered the same value twice in a row, the second time it would be accepted because no validation would occur. The disadvantage of this method is that, on the second input page, if the user was attempting to change a value that he/she still wanted validated, this would not occur.

A more advanced idea would be to have some way of turning off validation in a parameter-specific way (e.g. checkboxes for each parameters), but the effort level would be higher.

Planning on implementing the simple idea above unless there is a good reason to proceed with something else. Just chime in @Amy-Xu and/or @MattHJensen

talumbau:
Resolves #118

### DIFF
--- a/templates/taxbrain/input_form.html
+++ b/templates/taxbrain/input_form.html
@@ -43,6 +43,7 @@
 
     <form class="inputs-form" method="post" action="/taxbrain/?start_year={{start_year}}">
       {% csrf_token %}
+      <input type="hidden" name="has_errors" value="{{ has_errors }}" />
       <div class="inputs-sidebar">
         <label class="">Start Year:
           <select id="start-year-select" class="form-control pull-right">
--- a/webapp/apps/taxbrain/views.py
+++ b/webapp/apps/taxbrain/views.py
@@ -4,6 +4,7 @@ import json
 import taxcalc
 import dropq
 import datetime
+import logging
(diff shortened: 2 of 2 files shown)


talumbau:
In order for the user to know that he/she may submit with validation errors, I changed the error message that one will see upon first submission of values with a validation error:

screen shot 2016-01-07 at 1 26 08 pm

cc @Amy-Xu @MattHJensen

MattHJensen:
@talumbau, that seems pretty good to me. 

Will the user have to retype the numbers on the new page? If not, I'd suggest "submitted again from this page" rather than "submitted again on this page".

talumbau:
Oh yes I see. The user does not have to retype the numbers on the new page. So I will change to "submitted again from this page"

PR #125, opened 2016-01-11
Fix errors as warnings

talumbau:


talumbau:
The ModelForm by default will not allow a value to be saved if it
  doesn't pass the validation. So, we keep track of these values and
  then force them on to the model by grabbing the user-entered value out
  of the POST.

PR #134, opened 2016-01-21
PR: Table Builder

mistakevin:
refers https://github.com/OpenSourcePolicyCenter/webapp-public/issues/35
@talumbau  Still a work in progress, but you can pull to see functionality

screen shot 2016-01-21 at 2 29 18 pm
screen shot 2016-01-21 at 2 29 35 pm


MattHJensen:
Looks cool!

Will the table drill down and column selector dynamically change as different options are selected? For instance, if someone chose difference table rather than diagnostic table, there would be no 'user' vs 'base' plan option and the columns would be different.

mistakevin:
@MattHJensen @talumbau 
Feature complete.

Features:
* Table Builder Options
* Column Selector
  * Auto updates with current columns to choose from based on table type selected
* Exports:
  * CSV, EXCEL
  * Print View
  * Copy to Clipboard

Awaiting any design guidance, otherwise complete. Here is the current design.
screen shot 2016-01-22 at 12 27 29 pm
screen shot 2016-01-22 at 12 28 05 pm

MattHJensen:
@feenberg, could you take a look at this post from Kevin? https://github.com/OpenSourcePolicyCenter/webapp-public/pull/134#issuecomment-174003736

MattHJensen:
@mistakevin, just highlighting these questions again: https://github.com/OpenSourcePolicyCenter/webapp-public/pull/134#issuecomment-173967609

@feenberg, could you give us your feedback on Kevin's work? https://github.com/OpenSourcePolicyCenter/webapp-public/pull/134#issuecomment-174003736

mistakevin:
@MattHJensen Yes, the options change based on whether you select diagnostic vs difference. The User vs Plan is replaced with a "Payroll (+) Income" selector

MattHJensen:
Very cool. @mistakevin, can we play with this on a test server? 
cc @talumbau

talumbau:
yes, I will deploy to ospc-taxes5 before end of day.

MattHJensen:
Great. Thanks @mistakevin and @talumbau

MattHJensen:
Some comments from @feenberg after I shared the live site. I'll annotate these in a bit:

I don't have Word on this computer, but I was able to cut and paste a table into Open Office 2010 easily and with good results. That is a very important feature. It is true, however, that some of the column headings are so long that it rather distorts the table. The distortion is actually done by OO, the display on the webpage uses multiple lines but OO doesn't. For example, "Combined Payroll and Individual Income Tax Liabilities" makes for a very wide cell and a lot of wasted space. It even looks bad. Can we reduce that to (perhaps) "Payroll + Income Tax"? It seems that editing the column header in OO doesn't change the column width. Have you tried this with Word? How does it work?

Select columns is very clear and well done.

"Billions of Dollars" could be shortened to "Billions".

Does "Copy" put the table in the clipboard? I was able to load the CSV version into Open Office, but not the Excel version. I suppose that is ok.

There is a question about how the data should be arranged in the spreadsheets. There is an argument that there shouldn't be multiple tables just rows for years, then all the variables spread out across the columns. That would be good for machine consumption, but awkward to read because of the extensive right to left scrolling.

I wasn't able to actually run the model from this page - is that expected for the working version?

I still think there should be a line at the bottom like this:

 Source: http://www.ospc/taxbrain/451

As it is, the Taxbrain link won't be part of a printed table.

I don't actually understand much of the page.

Shouldn't it be possible to rerun with changed parameters? I don't see a "submit" button.

What is "For Year" about? Do I select a year to see the output for that year? (Are all years calculated, but only the requested year shown?). Can I ask for "All years"?

I don't understand anything underneath "For Year". "Diagnostic/Difference" appears to be something related to showing aggregate values versus aggregate difference from base, but the word "Diagnostic" has a different connotation.

"Base/User" is a radio buttion, but I have no idea what it does.

"Expanded Income/Adjusted Income" is clearer, but should have a longer explanation if one hovers. Is "Adjusted Income" AGI?

"Income Decimals"? Should be "Income Deciles". It could be labeled "Percentiles" even if we only offer a few selected percentiles. Do we think deciles are what we want to show? Wouldn't quintiles, then 80-90,90-95,95-99 and 99-100 be more interesting? Any idea why some row labels are split into 2 lines, while others are on a single line? Should we add a couple of pixels to the width of that box?

Overall, looks good but I still have requests.

dan

MattHJensen:
@mistakevin and @feenberg, here are some answers and annotations to Dan's questions

> "Billions of Dollars" could be shortened to "Billions".

Agreed

> I still think there should be a line at the bottom like this:
Source: http://www.ospc/taxbrain/451
As it is, the Taxbrain link won't be part of a printed table.

Agreed. @mistakevin, is there a way to incorporate a source in all forms of output? 

> Does "Copy" put the table in the clipboard? I was able to load the CSV version into Open Office, but not the Excel version. I suppose that is ok.

Yes it does put it in the clipboard. @mistakevin, is it easy to keep the table formatting when pasting into word? 

> What is "For Year" about? Do I select a year to see the output for that year? (Are all years calculated, but only the requested year shown?). Can I ask for "All years"?

"For Year" allows you to see the output for that year. All years are calculated ahead of time. I think we could add an "all years" feature if you think it would be helpful, but let's do it in a separate PR. 

> I don't understand anything underneath "For Year". "Diagnostic/Difference" appears to be something related to showing aggregate values versus aggregate difference from base, but the word "Diagnostic" has a different connotation.

@feenberg, these are different types of tables. Do you have suggestions for better names? I agree that these aren't very informative. 

> I wasn't able to actually run the model from this page - is that expected for the working version?

@mistakevin, could you add back the "edit parameters" button? 

> "Expanded Income/Adjusted Income" is clearer, but should have a longer explanation if one hovers. Is "Adjusted Income" AGI?

@mistakevin, how hard would it be to have hover over boxes to describe some of these table options? Alternatively, we could have footnotes at the bottom of the tables that change depending on which options you are seeinng.

> "Base/User" is a radio buttion, but I have no idea what it does.

Let's call this: "Baseline" and "Reform". 

Let's swap these in widely. 

> "Income Decimals"? Should be "Income Deciles". It could be labeled "Percentiles" even if we only offer a few selected percentiles. Do we think deciles are what we want to show? Wouldn't quintiles, then 80-90,90-95,95-99 and 99-100 be more interesting? 

Let's stick with deciles for this PR, but they should be called deciles rather than decimals. 

> Any idea why some row labels are split into 2 lines, while others are on a single line? Should we add a couple of pixels to the width of that box?

@mistakevin, what do you think?

mistakevin:
- [ ] remove "of Dollars"
- [ ] add source link to exports
- [ ] work on pasting into word
- [ ] replace Edit Parameters button
- [ ] add tooltips
- [ ] Base/User to Baseline/Reform
- [ ] single row row-labels

talumbau:
@mistakevin hmm... we have tooltips for policy parameters from the JSON file in taxcalc. In this case, I wonder if the tooltip content could come from a flatblock so that @MattHJensen could edit it as desired. Matt, what do you think of that idea?

mistakevin:
@talumbau I see them now. I'll plug those in

mistakevin:
@talumbau nevermind. I don't see them. Where are they?

mistakevin:
@talumbau @MattHJensen All but one of the checkboxes above are complete. As for the Word pasting, I don't have Word that I can test it on explicitly, but internet posts suggest if you create a table first and paste into it, it should fill in the cells correctly. This does work on Apple Pages, but not on Google Spreadsheets (at least not that way)

MattHJensen:
Thanks @mistakevin! @talumbau, can you push to the live site again and then @feenberg and I will make another review pass?

talumbau:
@MattHJensen @feenberg the latest changes from @mistakevin are up on the test app.

MattHJensen:
Could we use "Baseline" and "Reform" in table titles as well?
image

Until we have better names, what do you think about calling the two types of tables, "Diagnostic" Table and "Difference" Table? (With the quotes and inclusion of the word table)

feenberg:
On Sat, 30 Jan 2016, Matt Jensen wrote:

> 
> Could we use "Baseline" and "Reform" in table titles as well?
> image

We have decided against PlanX/PlanY?

> 
> Until we have better names, what do you think about calling the two types of tables,
> "Diagnostic" Table and "Difference" Table? (With the quotes and inclusion of the word
> table)

Aggregates/Deltas?

dan

MattHJensen:
@mistakevin there is some buggy behavior with the table selector. 

For example, if I have "difference" selected, like this 
image

And then click "income", I get this:

image

MattHJensen:
> Aggregates/Deltas?

@feenberg, the problem with this is that we might later end up displaying deltas of the aggregate table.

MattHJensen:
> We have decided against PlanX/PlanY?

@feenberg,I've found that Plan X and Plan Y are not very familiar to people who haven't worked in this domain. Even economists and policy analysts who haven't worked on this type of model don't understand these terms naturally. I've had more success with "baseline" and "reform". Are there disadvantages to using those terms that I'm missing?

mistakevin:
@MattHJensen Yep, it was a typo. Should be fixed now

mistakevin:
@talumbau ^^

feenberg:
On Mon, 1 Feb 2016, Matt Jensen wrote:

>
>       We have decided against PlanX/PlanY?
> 
> @feenberg,I've found that Plan X and Plan Y are not very familiar to people who
> haven't worked in this domain. Even economists and policy analysts who haven't worked
> on this type of model don't understand these terms naturally. I've had more success
> with "baseline" and "reform". Are there disadvantages to using those terms that I'm
> missing?
>

I rather agree about most potential users, I was thinking that it might be 
important to appeal to insiders, and using their jargon might be important 
to give the impression that we were joining their religion. I don't have a 
strong opinion.

dan

martinholmer:
Dan @feenberg and @MattHJensen,

I'm not a tax-policy insider, but I am an experienced policy analyst.
I've never heard anybody use the planX/planY terminology.  Everyone
I've talked to over the years would immediately understand "baseline"
and "reform" (although I would have a preference for "current law" rather
than "baseline", but that could just be me).

MattHJensen:
"current law" and "reform" sound good to me.

feenberg:
On Mon, 1 Feb 2016, Martin Holmer wrote:

> 
> Dan @feenberg and @MattHJensen,
> 
> I'm not a tax-policy insider, but I am an experienced policy analyst.
> I've never heard anybody use the planX/planY terminology. Everyone
> I've talked to over the years would immediately understand "baseline"
> and "reform" (although I would have a preference for "current law" rather
> than "baseline", but that could just be me).

It seems to be JTC and OTA jargon. The objection to "reform" is that it 
prejudges the plan, since "reform" implies improvement. But I don't really 
care - just wanted to make sure it was a non-random choice.

dan

feenberg:
On Mon, 1 Feb 2016, Matt Jensen wrote:

> 
> "current law" and "reform" sound good to me.

Yes, "current law" is better than "baseline".

dan

martinholmer:
Dan @feenberg said:
> It seems to be JTC and OTA jargon. The objection to "reform" is that it prejudges the plan, since "reform" implies improvement. But I don't really care - just wanted to make sure it was a non-random choice.

That's interesting.  In forty years of policy analysis work I've never heard the notion that a "reform implies improvement".  In my experience, a reform is simply a change from current-law policy.

Dan also said:
> Yes, "current law" is better than "baseline".

Do we have a consensus on "current law" and "reform" (instead of planX and planY)?

cc @MattHJensen

MattHJensen:
Yea, I've heard plan X plan Y in the tax crowd but agree that current law and baseline are clearer. Consensus reached. Thanks @martinholmer and @feenberg.

mistakevin:
@talumbau Updated to Current Law / Reform

MattHJensen:
Could we all do another review pass on the test site?

MattHJensen:
@mistakevin, when I first go to an output page, either by running a simulation or linking to it, the output tables do not populate: 
image

If I then select a different year, they will populate.

mistakevin:
@talumbau @MattHJensen "First Year" bug squashed.

MattHJensen:
@talumbau, could you push this to the test server? 

If everything looks good there and @mistakevin gets the tests fixed, let's merge!

We are also probably due for another tax-calculator release since there were some bug fixes recently and a new baseline.

MattHJensen:
Let's call the top categories "Diagnostic Table" and "Difference Table". After that is changed and the tests are passing, this looks ready to go. It's very cool. Thanks @mistakevin. 

Follow up questions: 

Is there something else that needs to happen on the diagnostic table side to get the agi tables working? 

Also, @mistakevin, did you end up linking the tool tips to flat blocks?

mistakevin:
@MattHJensen We put the tooltips in [here](https://github.com/OpenSourcePolicyCenter/webapp-public/blob/270fbb4bb93803699387f2a18086b6f88df38d1e/webapp/apps/taxbrain/constants.py) for now

MattHJensen:
> @MattHJensen We put the tooltips in here for now

Thanks @mistakevin.

Issue #137, opened 2016-01-22
Remove FEI parameter and note for payroll tax


Amy-Xu:
@talumbau found two things when I was doing my other task:

![capture](https://cloud.githubusercontent.com/assets/10549597/12520874/0825cffe-c115-11e5-8c27-b01143dd0da3.PNG)
This parameter won't work right due to lack of data. So we should remove this one from input page for now.

![capture1](https://cloud.githubusercontent.com/assets/10549597/12520896/27c39b52-c115-11e5-9370-fda59651a4d9.PNG)
This 'no payroll revenue' should also be taken off as we have payroll reported now.

Thank you! 
cc @MattHJensen

talumbau:
closed via #142

PR #142, opened 2016-01-27
Put Max. For. Earned Income in 'coming soon'

talumbau:


talumbau:
This resolves #137 

Maximum foreign earned income is now 'coming soon' so that the user can't modify it:

screen shot 2016-01-27 at 11 40 52 am

I also changed the flatblock on ospc.org so that it does not say the text "We do not currently report payroll tax revenue".

PR #157, opened 2016-02-13
update label for frisch elasticity

jdebacker:
This PR simply changes the label for the box where the Frisch elasticity parameters used in OG-USA is entered.  The new name is more descriptive.


talumbau:
Well, it didn't take long for this duplicate info to cause confusion. Sorry to make this confusing, but I think the best place to update this information is in the JSON file over on OGUSA:

https://github.com/open-source-economics/OG-USA/blob/master/Python/ogusa/parameters_metadata.json

Each time we version `ogusa`, whoever is doing the release will run a script that will generate a new version of the JSON file over here, check it in to webapp and then update the site. I *really, really* did not want to have this duplicate info (for obvious reasons) but I just couldn't see any other way around it. Here is the story:

Prior to adding this JSON file (and a couple of other JSON files for `ogusa`-relevant information), we would get this information by importing the `ogusa` package and then calling the proper functions on the package to get the metadata we need for various parameters. This is the correct way to do things because it means that the information is only specified in one place. However, recently it became clear that we can no longer reliably deploy to Heroku with the `ogusa` package. The issue is that Heroku has a maximum size of 300 MB for a deployed site (called a "slug"). This is typically not a problem because most websites don't have significant compiled assets like `numpy`, `scipy` and so forth. The `ogusa` package in particular has dependencies of `scipy` (which includes `numpy`) and the `matplotlib` package. This ends up taking ~150 MB just by itself. I was able to do some tricks to handle certain cases (for example, you can install the package, then actually remove numpy and as long as you don't call anything that uses numpy you are fine), but it turns out that in order to actually get data out of the `parameters` module in `ogusa`, we actually need a call in `scipy.optimize`:

https://github.com/open-source-economics/OG-USA/blob/master/Python/ogusa/elliptical_u_est.py#L163

So, the only way to get the proper parameter values out of `ogusa` is to have all of the (quite large) components installed. This is not possible with Heroku, so I had no choice but to dump out the relevant info for version 0.5 and then install it from files over on the webapp.

This experience has led me to conclude that we have perhaps "outgrown" Heroku as a service, but the effort required to leave Heroku completely would probably be somewhat significant.

So, short story: please update this info over on OGUSA repo and we will try to keep it up to date with OGUSA - or, open a PR on each repo so that we can stay synced up that way. My goal though, would be for OGUSA contributors to be oblivious to this issue.

talumbau:
So, just to be clear, Heroku is the "platform service" that hosts the site itself. The worker nodes run on AWS and obviously have `ogusa` installed with whatever is the latest released version.

jdebacker:
@talumbau, Thanks for this.  I had no idea, obviously.  I opened the same PR in the OG-USA repo.  I don't know if you want to close this one or just merge both so that they files are in sync across the repos.

talumbau:
The test failure is spurious because the source branch of the PR is not on this repo. This is just an update to a JSON file. merging.

PR #168, opened 2016-03-01
Use mock to test submitting job on TaxBrain.

talumbau:
Here is the initial infrastructure to more thoroughly test TaxBrain. We use `requests_mock` for simulating HTTP responses. Instead of full dependency injection, we monkey patch the `dropq_compute` object when testing. This seemed to be a reasonable way to get started for now, although it's not the cleanest. We will likely want to add a couple more tests here before merging.


talumbau:
This is complete enough to be of use. In particular, it will be needed to add tests for #170 and #171

PR #185, opened 2016-03-11
Updated Bootstrap, using bottom offset for aside

mistakevin:
Was a Bootstrap bugfix https://github.com/twbs/bootstrap/issues/15032. Updated and applied.
Resolves https://github.com/OpenSourcePolicyCenter/webapp-public/issues/184


talumbau:
Hmm.. @mistakevin I'm getting the previous behavior of the sidebar "jumping up":

screen shot 2016-03-12 at 2 18 55 pm

I'm running with `DEV_DEBUG=True` so all static content is served from `/static/` so I feel confident I'm using the updated javascript. Anything I can do to help debug this?

talumbau:
upon further review, it appears that bootstrap 3.3.6 does indeed resolve this issue on Chrome, Safari, and Firefox. I'm not sure how I got the bad result above. I'm think we should merge this and deploy since I can no longer get it to fail as before. This leads me to conclude I was doing something wrong and that the fix here is correct. cc @jdebacker @MattHJensen

PR #188, opened 2016-03-11
AMTR -> 1 - AMTR. Small text changes. 

MattHJensen:
I missed a couple of places where AMTR should be 1 - AMTR. These are small text changes. 

cc @talumbau @mistakevin


talumbau:
these are minor HTML changes only. I'm fine with merging w/o running through Travis CI. the failure is simply because the originating branch is not this repo but rather Matt's fork. PRs set up in this way don't get all of the environment variables that Travis CI needs to run the tests (for security purposes). Going forward with the merge.

Issue #182, opened 2016-03-11
re-runnning results leads to 500 error


MattHJensen:
I get a dynamic result
http://www.ospc.org/dynamic/behavior_results/21/

Then I click "edit parameters" 
http://www.ospc.org/dynamic/behavioral/edit/21/?start_year=2016

Then I try to rerun the simulation:
http://www.ospc.org/dynamic/behavioral/21/?start_year=2016

And I get a 500 internal server error. 

This problem does not exist for the static model. I am not sure if it does for the macro elast model; I haven't checked that yet. 

cc @talumbau @mistakevin

talumbau:
resolved via https://github.com/OpenSourcePolicyCenter/webapp-public/pull/190

PR #190, opened 2016-03-12
Fix 182

talumbau:


talumbau:
Resolves #182 

This is a one-line change in actual web app code. The only other changes are:

- A fairly involved test case to catch the bug caught in #182 
- changing the tests to run with `py.test` (because it's just better)
- some insignificant commit history that got it while troubleshooting a Heroku problem.

cc @MattHJensen

Issue #202, opened 2016-03-17
EITC parameters


Amy-Xu:
```
reform = {2016:{"_EITC_rt":[[0.08415, 0.374, 0.44, 0.495]],
                "_EITC_prt": [[0.08415, 0.17578, 0.23166, 0.23166]]}}

```

It seems this reform has inconsistent results between taxcalc and TaxBrain. Using taxcalc, the difference between baseline and reform is within 2 billions, but on TaxBrain the difference is roughly between 70 - 80 billions, running separately or combined together.

@talumbau @MattHJensen @martinholmer

talumbau:
closed via #203

PR #203, opened 2016-03-17
fix packaging up rates and propagating forward

talumbau:
- don't round to nearest integer on rates


talumbau:
Resolves #202

martinholmer:
When is the bug fix in pull request #203 going to be available in the public version of TaxBrain?

@talumbau @MattHJensen

talumbau:
@martinholmer it's live on the site right now.

PR #209, opened 2016-03-24
Only read JSON response after checking status code

talumbau:
- Fixes a bug that would crash the app if attempting to submit to a node
  that is misbehaving. We only look at the contents of the repsonse if
  we get a 200 from the worker node


talumbau:
This fixes a bug that was introduced on March 9 by this commit:

https://github.com/OpenSourcePolicyCenter/webapp-public/commit/183894cddfea0bb003305cce04476fe810ae399b

where the JSON response was grabbed prior to checking the status code.

PR #217, opened 2016-04-05
Ogusa round robin

talumbau:


talumbau:
This PR ensures that OGUSA jobs are assigned in a round-robin fashion across all designated OGUSA nodes.

Issue #238, opened 2016-05-09
Implausible delayed-reform-implementation results?


martinholmer:
When I enter `*,*,250000` in the text box for Additional Medicare threshold for Single filers (when TaxBrain start year is 2016), I expect to see no payroll tax revenue change in 2016 and 2017, and then beginning in 2018, I expect to see a modest decline in payroll tax revenue.  The [TaxBrain results](http://www.ospc.org/taxbrain/3646/) do show a modest decline in payroll tax revenue, but those declines start in 2017, not 2018.  Did I do something wrong or is there a problem with the new TaxBrain?

@talumbau

talumbau:
I spent some time trying to think of a general way to capture all possible inputs combinations with the wildcard and then test all of them at once (the current pattern of 'fix, bug report, repeat' seems pretty inefficient). I didn't really come up with anything good. For example, I can pretty easily create a whole bunch of potential input fields. There is a function called `default_policy` in the web app that is very similar to `default_data` in Tax-Calculator, although its return value is special.

```
        dd = default_policy(2016)
        all_names = []
        for k, v in dd.items():
            for col in v.col_fields:
                all_names.append(col.id)
        num_per_combo = len(all_names) - 2
        some_combos = [a_combo for a_combo in itertools.combinations(all_names, num_per_combo)]
```

 By doing some additional processing, I end up with `all_names` which is every field on TaxBrain, including each multi-valued parameter broken down into individual fields (so `_AMED_thd` becomes `AMED_thd_0`, `AMED_thd_1, ..., `AMED_thd_3`). Then, to get a manageable number of combinations, I grab something like all combinations of size 178 from the set of 180 fields. This ends up with about 18,000 combinations.

The hard part is then, for each one, coming up the expected answer. This is probably the wrong approach. @martinholmer can you provide me a list of inputs you intend to use with the wildcard behavior? I could try these out locally and then just push all of the fixes together if there is a problem.

For this particular filed issue, it's resolved with #240

talumbau:
#240 is live on TaxBrain so this issue should be resolved.

martinholmer:
T.J. (@talumbau) said:
> Can you provide me a list of inputs you intend to use with the wildcard behavior? I could try these out locally and then just push all of the fixes together if there is a problem.

[...]

PR #240, opened 2016-05-10
Fix multi-value params with multiple wildcards

talumbau:
- indexing was incorrect in plugging in expanded defaults
- add test for multiple wildcards in a multi-valued parameter


talumbau:
resolves #238

PR #243, opened 2016-05-11
Dynamic Progress Bar

### LINKED ISSUE #241: Waiting screens for two dynamic simulations should be the same as microsim
For partial equilibrium and macroeconomic elasticity dynamic simulations, we use the same "old style" waiting page that was replaced by @mistakevin's very nice loading screen. These two waiting screens should be changed to be like the microsim waiting screen. The OGUSA simulation flow should be the same (since we are just sending an email upon job submission)

mistakevin:
resolves #241 

Only frontend. WORK IN PROGRESS!


talumbau:
Added backend changes so that is probably good to go now.

mistakevin:
Working for me

talumbau:
ok great. Merging.

PR #250, opened 2016-05-19
Replaced form style

mistakevin:
resolves https://github.com/OpenSourcePolicyCenter/webapp-public/issues/249


talumbau:
Looks good. Merging and then will deploy.

PR #258, opened 2016-05-27
Accept wildcards on fields that validate

### LINKED ISSUE #257: TaxBrain "500 Internal Server Error!" bug for tax bracket reforms
TaxBrain crashes when a user tries to specify a change in income tax brackets that are delayed one or more years after the TaxBrain start year.

For example, the 500 error is generated when start year is 2016 and the following is entered into the `Single` box under the title `Personal income tax bracket (upper threshold) 1`:
```
*,*,15000
```
This notation is specifying an increase in the top of the first income tax bracket beginning in 2018.

Trying to raise other brackets for single or other filing-unit types also produces the 500 error.

@talumbau @MattHJensen

talumbau:
- Resolves #257
- Enhance form logic to handle wildcard entry prior to field validation

### DIFF
--- a/webapp/apps/taxbrain/forms.py
+++ b/webapp/apps/taxbrain/forms.py
@@ -4,12 +4,56 @@ from django.utils.translation import ugettext_lazy as _
 
 from .models import TaxSaveInputs
 from .helpers import (TaxCalcField, TaxCalcParam, default_policy, is_number,
-                      int_to_nth, is_string, string_to_float_array)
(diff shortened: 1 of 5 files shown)


martinholmer:
T.J.,  Thanks for all the changes and new tests in pull request #258.  When do expect these changes to be incorporated in the publicly-available version of TaxBrain?

talumbau:
The latest push addresses validation where both parameters have wildcards and inflation must be taken into account on both (and respected if turned off). Example:

screen shot 2016-05-27 at 1 13 49 pm

PR #278, opened 2016-07-11
Added start year modal

### LINKED ISSUE #273: change start year UI?
When someone changes the start year, a small text notice is displayed:

image

I believe some users may not see this and are then surprised when the tables start from the default start year instead of the start year they wanted to set. 

@talumbau and @feenberg , what do you think about presenting a pop up that takes over the screen instead? It would need to say something like: "Changing the default start year will clear any tax law inputs you have made."..."Ok, go for it" or  "No, I want to keep the current start year".

mistakevin:
resolves #273 

screen shot 2016-07-11 at 9 59 51 am


talumbau:
this worked well for me when I just loaded the page, but if I click on a section on the sidebar (e.g. "Standard Deduction") and then select a different year. The page gets darker but the modal pop-up is sort of "cut off". Like this:

screen shot 2016-07-12 at 5 02 31 pm

then I'm unable to get the window to close so I have to refresh the page. 

cc @mistakevin @MattHJensen

mistakevin:
Nice find. Updated @talumbau

Issue #314, opened 2016-08-25
Re-activate "Growth" section on TaxBrain


MattHJensen:
I think the "Growth" section ought to be renamed as well. My proposal is "Baseline Growth".

talumbau:
closed via #317

PR #317, opened 2016-08-26
Add Growth Assumptions parameters back to TaxBrain

talumbau:
- change name to "Baseline Growth Assumptions"


talumbau:
resolves #314

PR #319, opened 2016-09-06
Ccc latest

talumbau:


PeterDSteinberg:
Last commit has changes to `templates/taxbrain/header.html` to allow `is_btax`: True/False to be passed in, changing the logo at the top of the header that way.  Here is the results.html look now:

screen shot 2016-09-08 at 6 31 19 am

Here is the `not_ready.html` view:

screen shot 2016-09-08 at 6 30 30 am

PeterDSteinberg:
last commit fixes #329

PeterDSteinberg:
last commit fixes #325

talumbau:
Hmm... looks like master needs to be merged in here. @PeterDSteinberg can you do this?

talumbau:
I think this is ready to go on OSPC.org. The "landing page" is hidden for now, so users can only link to TaxBrain. We can modify with a simple follow-up PR if everything looks good on ospc.org. Merging.

PR #338, opened 2016-09-12
Add flag to increment worker nodes counter

talumbau:
- For BTax workloads, we don't need to increment the counter
   that keeps track of of offset in the worker node pool

 - depends on #319


talumbau:
resolves #334 / #336

PR #346, opened 2016-09-30
JSON based input page taxbrain/json

### LINKED ISSUE #312: Add TaxBrain ability to specify run using JSON reform file
The discussion beginning around August 1, 2016, concerning [Tax-Calculator issue #429](https://github.com/open-source-economics/Tax-Calculator/issues/429#issuecomment-240692920) has concluded that we want to enhance TaxBrain to be able to handle an alternative way of specifying policy reforms.  In addition to the text-box and check-box approach currently used by TaxBrain, we want to give users the option to bypass that approach and specify a reform using a JSON file as can be done when using the inctax.py interface to Tax-Calculator.

This TaxBrain enhancement would use existing Tax-Calculator capabilities, so this enhancement requires no changes to Tax-Calculator.  There is an example of a JSON reform file in the test_incometaxio.py file [REFORM_CONTENTS string](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_incometaxio.py#L173-L208).  And there is a [Policy class static method](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/policy.py#L174-L209) that reads a JSON reform file and returns a dictionary that can be passed to the Policy class implement_reform method.

Here are some of the features this new TaxBrain capability should probably have:

(1) the ability to specify a local JSON reform file for uploading to TaxBrain;

(2) syntax checking that indicates illegal JSON;

(3) a way of saving the reform file associated with TaxBrain run results;

(4) a way of editing a saved reform file to generate a new TaxBrain run;

(5) a way of downloading a saved JSON reform file.

I don't use TaxBrain often, so I don't have much experience with the work flow.  I'm hoping that others with more TaxBrain experience will revise and/or add to this list of features.

@MattHJensen @talumbau @feenberg @Amy-Xu @GoFroggyRun @zrisher

talumbau:
- Error messages on failure to parse JSON
 - If valid JSON but unable to run microsim, present remote error on
   failed job page
 - Input boxes for Calculator, Growth, and Behavior

Resolves #312

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ install:
   - conda create -n webappdev python=$TRAVIS_PYTHON_VERSION pytest setuptools 
   - source activate webappdev
   - conda install -c ospc --file conda-requirements.txt
+  - conda update -c ospc --force taxcalc
   - pip install -r requirements.txt
   - pip install -r requirements_dev.txt
   - pip install pytest-django
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: newrelic-admin run-program waitress-serve --port=$PORT webapp.wsgi:application
+web: newrelic-admin run-program waitress-serve --max-request-body-size=2097152 --port=$PORT webapp.wsgi:application
--- a/static/js/btax.js
+++ b/static/js/btax.js
@@ -28,7 +28,7 @@ $.each(asset_yr_str, function(index, value){
 });
 
 $('#start-year-select').change(function() {
-  $('#current-year-link').attr('href', '/btax/?start_year=' + $(this).val());
+  $('#current-year-link').attr('href', '/ccc/?start_year=' + $(this).val());
   $('#current-year-alert').removeClass('hidden');
 });
 
--- a/static/js/filetb.js
+++ b/static/js/filetb.js
@@ -1,0 +1,88 @@
+// taxbrain.js
+
+//
+//  Input page
+//
+
+// sidebar - scrollspy
+$('body').scrollspy({
+  target: '.inputs-sidebar'
+});
+
+// Scroll to taxbrain form input areas, instead of jumping
+$('.inputs-form a, .btn-explore').click(function(event) {
+  if ($(this).attr('id') === 'current-year-link') {
+    return;
+  }
+  // ensure link is indeed a hash link on current page <<-- Zach's stuff
+  if (this.host == location.host && this.pathname == location.pathname) {
+    event.preventDefault();
+    var target = this.hash;
+    if (this.href != '#') {
+      $('body, html').animate({
+        scrollTop: $(target).offset().top
+      }, 700);
+    }
+  }
+
+});
+
+//
+//  Results page
+//
+
+// Select behavior for dropdowns
+$(".dropdown-select .dropdown-menu li a").click(function(){
+  selection = $(this);
+  dropdown_label = $(this).parents('.dropdown-select').find('.dropdown-toggle')
+  dropdown_label.html(selection.text() + ' <span class="caret"></span>');
+});
+
+// Switch behavior for dropdowns
+$(".data-switch .dropdown-menu li a").click(function(){
+  switch_key = $(this).text().trim();
+
+  /*
+  //Without fade-out-in
+  $( ".data-switchable" ).each(function(){
+    $this = $(this)
+    switch_value = $this.data("switch-options")[switch_key]
+    $this.text(switch_value)
+  })
+  */
+
+  // With fade-out-in
+  $('.data-switchable').fadeOut('fast',function(){
+    $this = $(this)
+    switch_value = $this.data("switch-options")[switch_key]
+    $this.text(switch_value).fadeIn('fast');
+  });
+
+});
+
+$('input[name="growth_choice"]').click(function () {
+  if ($(this).val() === 'factor_target') {
+    $('#id_factor_adjustment').val('');
+  } else {
+    $('#id_factor_target').val('');
+  }
+});
+
+$('#id_factor_adjustment, #id_factor_target').focus(function() {
+  if ($(this).attr('id') === 'id_factor_target') {
+    $('input[type="radio"][value="factor_target"]').click();
+  } else {
+    $('input[type="radio"][value="factor_adjustment"]').click();
+  }
+});
+
+var currentYear = $('#start-year-select').val();
+$('#start-year-select').change(function(e) {
+  $('#current-year-link').attr('href', '/taxbrain/file/?start_year=' + $(this).val());
+  $('#current-year-modal').modal('show');
+});
+
+$('#current-year-modal').on('hide.bs.modal', function (e) {
+  $('#start-year-select option').removeAttr("selected");
+  $('#start-year-select option[value="' + currentYear + '"]').attr("selected", "selected");
+});
--- a/staticfiles/js/btax.js
+++ b/staticfiles/js/btax.js
@@ -28,7 +28,7 @@ $.each(asset_yr_str, function(index, value){
 });
 
 $('#start-year-select').change(function() {
-  $('#current-year-link').attr('href', '/btax/?start_year=' + $(this).val());
+  $('#current-year-link').attr('href', '/ccc/?start_year=' + $(this).val());
   $('#current-year-alert').removeClass('hidden');
 });
 
--- a/staticfiles/js/filetb.js
+++ b/staticfiles/js/filetb.js
@@ -1,0 +1,88 @@
+// taxbrain.js
+
+//
+//  Input page
+//
+
+// sidebar - scrollspy
+$('body').scrollspy({
+  target: '.inputs-sidebar'
+});
+
+// Scroll to taxbrain form input areas, instead of jumping
+$('.inputs-form a, .btn-explore').click(function(event) {
+  if ($(this).attr('id') === 'current-year-link') {
+    return;
+  }
+  // ensure link is indeed a hash link on current page <<-- Zach's stuff
+  if (this.host == location.host && this.pathname == location.pathname) {
+    event.preventDefault();
+    var target = this.hash;
+    if (this.href != '#') {
+      $('body, html').animate({
+        scrollTop: $(target).offset().top
+      }, 700);
+    }
+  }
+
+});
+
+//
+//  Results page
+//
+
+// Select behavior for dropdowns
+$(".dropdown-select .dropdown-menu li a").click(function(){
+  selection = $(this);
+  dropdown_label = $(this).parents('.dropdown-select').find('.dropdown-toggle')
+  dropdown_label.html(selection.text() + ' <span class="caret"></span>');
+});
+
+// Switch behavior for dropdowns
+$(".data-switch .dropdown-menu li a").click(function(){
+  switch_key = $(this).text().trim();
+
+  /*
+  //Without fade-out-in
+  $( ".data-switchable" ).each(function(){
+    $this = $(this)
+    switch_value = $this.data("switch-options")[switch_key]
+    $this.text(switch_value)
+  })
+  */
+
+  // With fade-out-in
+  $('.data-switchable').fadeOut('fast',function(){
+    $this = $(this)
+    switch_value = $this.data("switch-options")[switch_key]
+    $this.text(switch_value).fadeIn('fast');
+  });
+
+});
+
+$('input[name="growth_choice"]').click(function () {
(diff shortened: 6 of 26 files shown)


talumbau:
This PR should be a reasonable start on JSON input. It has the following features:

 - On failure to parse entered JSON, the JSON parse error is presented to the user:

screen shot 2016-09-29 at 11 32 55 pm

and the text box still has whatever text was originally put in so you can fix it.

- On correctly parsed JSON but incorrect Tax-Calculator parameters, the error from the work node is propagated back to the web server and presented to the user:

screen shot 2016-09-29 at 11 35 10 pm

The testing of this new input feature has not been exhaustive, but we could deploy it either to the production site or the test app, if a "bumpy start" is OK. Thoughts @MattHJensen ?

MattHJensen:
@talumbau, let's deploy to the testing site first. I'm excited to try it out.

martinholmer:
@talumbau said about TaxBrain pull request #346:
> Resolves #312

Unfortunately, that is not the case.

Issue #312, which is entitled "Add TaxBrain ability to specify run using JSON reform file", is an enhancement request that contains this description of what a "JSON reform file" means:

> There is an example of a JSON reform file in the test_incometaxio.py file [REFORM_CONTENTS string](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/tests/test_incometaxio.py#L173-L208). And there is a Policy class static method that reads a JSON reform file and returns a dictionary that can be passed to the Policy class implement_reform method.

The above link to the REFORM_CONTENTS string contains the following example of a "JSON reform file":
```
// Example of a reform suitable for use as an optional IncomeTaxIO reform file.
// This JSON file can contain any number of trailing //-style comments, which
// will be removed before the contents are converted from JSON to a dictionary.
// The primary keys are policy parameters and secondary keys are years.
// Both the primary and secondary key values must be enclosed in quotes (").
// Boolean variables are specified as true or false (no quotes; all lowercase).
{
    "_AMT_tthd": // AMT taxinc threshold separating the two AMT tax brackets
    {"2015": [200000],
     "2017": [300000]
    },
    "_EITC_c": // maximum EITC amount by number of qualifying kids (0,1,2,3+)
    {"2016": [[ 900, 5000,  8000,  9000]],
     "2019": [[1200, 7000, 10000, 12000]]
    },
    "_II_em": // personal exemption amount (see indexing changes below)
    {"2016": [6000],
     "2018": [7500],
     "2020": [9000]
    },
    "_II_em_cpi": // personal exemption amount indexing status
    {"2016": false, // values in future years are same as this year value
     "2018": true   // values in future years indexed with this year as base
    },
    "_SS_Earnings_c": // social security (OASDI) maximum taxable earnings
    {"2016": [300000],
     "2018": [500000],
     "2020": [700000]
    },
    "_AMT_em_cpi": // AMT exemption amount indexing status
    {"2017": false, // values in future years are same as this year value
     "2020": true   // values in future years indexed with this year as base
    }
}
```
Notice that this JSON reform file has the same structure as the TaxBrain GUI presents to TaxBrain users: the policy parameters are listed first with the timing and magnitude of reform changes listed for each parameter involved in the reform.

But pull request #346 does not implement the requested enhancement because it assumes the primary JSON keys are calendar years, not policy parameters.  This means that pull request #346 implements something that is not only incompatible with the JSON reform files used by the `inctax.py` and `simtax.py` command-line interfaces to the taxcalc package, but is inconsistent with the TaxBrain GUI.

Because there never was any discussion of the pros and cons of rejecting the enhancement request in issue #312, it would appear that the nature of pull request #346 is based on an oversight.  If so, then that oversight needs to be corrected.  If not, why was enhancement request #312 rejected without any discussion?

@MattHJensen

talumbau:
Ah, I see now. Yes, I did not dig in to the details here and was unaware the request was to support this new input format. A lot of the work of this PR needed to be done anyway to support the actual requested feature, so it seems like there is just a bit more to do. Seems reasonable to use [Policy.read_json_reform_file](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/policy.py#L175) to translate the JSON into a dictionary that can be passed to `Policy.implement_reform`. What about parameters for `Behavior` and `Growth` objects  that a user can currently specify via TaxBrain? In this PR, I just added additional boxes for specifying those parameters via JSON. Options that occur to me would be as follows:

 - take away those boxes and don't allow that kind of input for now
 - keep those input boxes and allow the standard JSON input mechanism for those parameters
 - keep those input boxes and create input process functions for Behavior and Growth parameters that are similar to `Policy.read_json_reform_file`.

Those options are in order of least effort, so the easiest way to get this feature in to production would be to do the first option. Thoughts @martinholmer @MattHJensen ?

martinholmer:
@talumbau said:
> What about parameters for Behavior and Growth objects that a user can currently specify via TaxBrain? In this PR, I just added additional boxes for specifying those parameters via JSON. Options that occur to me would be as follows:
>
> 1. take away those boxes and don't allow that kind of input for now
> 2. keep those input boxes and allow the standard JSON input mechanism for those parameters
> 3. keep those input boxes and create input process functions for Behavior and Growth parameters that are similar to Policy.read_json_reform_file.
>
> Those options are in order of least effort, so the easiest way to get this feature in to production would be to do the first option. 

I vote for option 1.  That way we can get experience with the JSON reform files while thinking about the best approach to Behavior and Growth parameters.

@MattHJensen

MattHJensen:
> I vote for option 1. That way we can get experience with the JSON reform files while thinking about the best approach to Behavior and Growth parameters.

👍

talumbau:
some progress here. If [this PR](https://github.com/open-source-economics/Tax-Calculator/pull/959) is merged, we can have error messages like this:

screen shot 2016-10-04 at 8 13 02 am

martinholmer:
@talumbau said:
> some progress here. If this PR is merged, we can have error messages like this ...

Because I was making other changes to the Policy class `read_json_reform_*` methods, I have incorporated your PR's ideas into [PR 958](https://github.com/open-source-economics/Tax-Calculator/pull/958) and merged 958.  So, now the Tax-Calculator master branch can support this kind of error message.

martinholmer:
@talumbau is proposing this kink of invalid-JSON error message in TaxBrain:
 


But I thought our prior discussion suggested that TaxBrain could offer a much more user-friendly approach (say, by putting the text box cursor at the first JSON error or by showing the line number and column number of the first JSON error).  I was envisioning TaxBrain catching JSON errors **before** calling one of the Policy.read_json_reform_* methods, so that this go-to-jsonlint.com message would never come up on TaxBrain.

@MattHJensen

talumbau:
If the expected input is only JSON, we can provide the error messages from the Python JSON parser to the browser. However, my reading of the issue is that the intention is to be able to provide non-JSON text. So, for example, an allowed input is supposed to be:

```
{
"_AMT_tthd": // AMT taxinc threshold separating the two AMT tax brackets
    {"2015": [200000],
     "2017": [300000]
    }
}

```

So, we just propagate this text to the backend and let `Policy.read_reform_file` read the text. Any error raised by this function is displayed. If I just use the error from the JSON parser on the above text, I would just get:

```
ValueError: No JSON object could be decoded
```

martinholmer:
@talumbau said:
> If the expected input is only JSON, we can provide the error messages from the Python JSON parser to the browser.  However, my reading of the issue is that the intention is to be able to provide non-JSON text.

All TaxBrain needs to do is replace the //-comments with a blank space (that is one Python statement) and then feed the resulting text into the JSON parser.  Won't that work?  The line and column numbers will be the same with or without the //-comments.

@MattHJensen

talumbau:
> All TaxBrain needs to do is replace each //-comment with a blank space (that can be accomplished using just one Python statement) and then feed the resulting text into the JSON parser. Won't that work? The line and column numbers will be the same with or without the //-comments.

This is a bad idea. 

It's almost never the right answer to create one's own language or data format to solve a problem. Similarly, it's almost never the right answer to *slightly modify* an existing data format to solve a problem, because what has actually happened is that one has actually created a new data format anyway. 

If this input format (JSON plus some ad-hoc commenting scheme) is useful for someone's purposes, I'm not going to object if it doesn't disrupt the use cases that I need to function properly, but this is not code that should go in this repository.

I have been down this route multiple times in multiple projects with extended lifetimes. Inevitably, this custom parsing behavior expands to include additional items, which are then solved with additional regular expressions, which never completely solve the problem because your users will find unintended ways to use the notation. You will then need to solve these additional use cases with yet more special cases and additional regular expressions.

If we go down this path, we will have the code to remove the `//` comments in two places. When this current scheme is found to be deficient in some way (which will happen), any change will have to occur in the web application and in `Policy.read_json_reform_text`. It's unlikely these will stay in sync unless someone remembers to change them both simultaneously.

In its current form, `Policy.read_json_reform_text` swallows the JSON parse error after the custom regular expressions processing. It would seem reasonable to augment this function as follows:

```
     try:
            reform_dict_raw = json.loads(json_without_comments)
     except ValueError as ve:
            msg = 'Policy reform text contains invalid JSON:'
            txt = ('\nTO FIND FIRST JSON SYNTAX ERROR,\n'
                   'COPY TEXT BETWEEN LINES AND '
                   'PASTE INTO BOX AT jsonlint.com\n')
            line = '----------------------------------------------------------'
            txt += line + '\n'
            txt += json_without_comments.strip() + '\n'
            txt += line + '\n'
            raise ValueError(msg + str(ve) + txt)
```

so basically we just capture the ValueError object from `json.loads` and add its contents to the final error message. This would allow the user to get the parse error. When I run my PR with a version of Tax-Calculator master with the single change above and this input (intentionally adding a single quote instead of a double quote):

```
{
"_AMT_tthd': // AMT taxinc threshold separating the two AMT tax brackets
    {"2015": [200000],
     "2017": [300000]
    }
}
```

I generate this output:

screen shot 2016-10-04 at 2 14 08 pm

I can create the PR on Tax-Calculator for this change.

talumbau:
Remaining items:

- [ ] "Edit Parameters" takes one back to JSON input page
- [ ] File upload widget

talumbau:
Actually, because the ultimate use case is to upload a file, we will have no "Edit Parameters" button on TaxBrain results that come from a JSON uploaded file. If we support an "Edit Parameters" button, it would defeat the purpose of having a file upload anyway, because we would take the user back to a text box full of the contents of the file, which is what this PR already supports and what we are deprecating in favor of file upload. Let me know if that doesn't sound right @MattHJensen

MattHJensen:
> Actually, because the ultimate use case is to upload a file, we will have no "Edit Parameters" button on TaxBrain results that come from a JSON uploaded file. If we support an "Edit Parameters" button, it would defeat the purpose of having a file upload anyway, because we would take the user back to a text box full of the contents of the file, which is what this PR already supports and what we are deprecating in favor of file upload. Let me know if that doesn't sound right @MattHJensen

I think it is important for the user to have some way to see what reform is modeled.

Our future plan is to make the reform visible from every output page, and not just the input page -- when that is complete, this won't be so important. 

But for now our only means of showing the user what parameters were changed is for the user to click "edit parameters". 

An ideal option (I think) would be to keep the text boxes and just have the json appear in the text box when the user uploads the file initially. Then, when he user clicks "edit parameters" it would makes sense to go back to the text box. 

Given that I don't know how tricky the above would be, another option would be to change "edit parameters" to "see parameters" for these runs and then just display the uploaded json as text on a pager.

@talumbau, what do you think about all this? @martinholmer might have some other ideas too since he's given a lot of thought to this feature.

talumbau:
Ah, good point. Yes, I like the idea of displaying the text that the user uploaded. That's the easiest to do right now. The "load file and display text in text box, then submit" story is a bit involved on the javascript side. So, if we just started with "upload file -> submit -> display results with file text also displayed" that would provide some good utility right now, correct? @MattHJensen @martinholmer

MattHJensen:
> So, if we just started with "upload file -> submit -> display results with file text also displayed" that would provide some good utility right now, correct? @MattHJensen @martinholmer

Yes, I think that is a useful first step.

talumbau:
@martinholmer @MattHJensen this is what the results page will include on TaxBrain runs generated from a file upload:

screen shot 2016-10-06 at 8 42 15 am

MattHJensen:
@talumbau, do we need to drop the "edit parameters" button too?

talumbau:
yep, I just did that after I uploaded the screenshot. Got ahead of myself!

martinholmer:
@talumbau said:
> [T]his is what the results page will include on TaxBrain runs generated from a file upload:

Looks fine assuming the uploaded file was formatted that way.  Thanks for all the work on this.

Issue #323, opened 2016-09-07
CCC UI: Info buttons on the depreciation rules


jdebacker:
Can we have info buttons on the depreciation rules table headers?  e.g. on ADS, GDS, etc.  These need some explanation.

If we can get these, I think we can omit the "Rate of bonus on n-year class property" and thus simplify and condense the table.

MattHJensen:
@jdebacker, do you want this text to reside in webapp-public as flatblocks or in B-Tax? (@talumbau/@brendancol, is one of these much easier?)

jdebacker:
I like having them in the json files in B-Tax and accessed on the web app via mousing over the (i) button.  I feel like that allows one to have a lot of information without cluttering the UI.  But happy to defer to these others who are more experienced.

talumbau:
I don't think we could do a flatblock for hover overs (not 100% sure but I think it would be tricky). Getting the text from a JSON file and displaying the text as a hover over for a button is what we currently do for TaxBrain so I would suggest we go that route here.

talumbau:
@jdebacker does the desired text already exist? If yes, can you point me to it?

jdebacker:
I think we could just pull from the "notes" filed in the `btax_defaults.json` file. @talumbau

talumbau:
Ok, I thought I understood then but it turns out I don't understand. In the `btax_defaults.json`, each item does have a notes section, but it's not clear to me which text should pop up when one hovers over the "ADS" text in the ADS column on the input page. Similar question for "GDS" and "Economic."

jdebacker:
Ah, I see, the notes read something like:

> "This parameter is an indicator for the use of the Alternative depreciation System (ADS) on 27.5-year class property."

So they are specific to each class of property.  I think for the hover-over at the top of the column, we just want to be a bit more general.  E.g.,

> "This parameter is an indicator for the use of the Alternative depreciation System (ADS)."

Can we create another variable for such text?

talumbau:
Yes that would be great.

jdebacker:
@talumbau  See [here](https://github.com/open-source-economics/B-Tax/pull/108)

talumbau:
[...]

PR #371, opened 2016-10-18
Add hover overs to column headings on CCC

talumbau:
- use text from 'hover' notes in btax default parameters


talumbau:
Resolves #323

PR #383, opened 2016-10-25
fix default output displayed

jdebacker:
This PR changes the default output displayed in the CCC output page to reflect the what is checked by the radio buttons as a default.  In particular, when the CCC output page is loaded, this PR ensures that METTRs by industry for the reform as displayed.


jdebacker:
@talumbau - could you review.  I think this fixes the issue with the default radio buttons and the order of the output variables listed (order should be mettr, metr, cc, deprec).

talumbau:
@jdebacker thanks for this fix. It took me a second to understand, but your change is definitely right. I'll merge and deploy as I consider this a high priority bug fix. Thanks!

PR #392, opened 2016-10-28
Fix /ccc row ordering issues: 385, 386, 387, 388

PeterDSteinberg:
This fixes issues related to row ordering on /ccc results tables and gets rid of the extra footer info.

Here is the new industry look:
screen shot 2016-10-28 at 12 16 01 pm

This shows the minor headings "Durable goods" and "Nondurable goods" inside "Manufacturing" section:
screen shot 2016-10-28 at 12 16 17 pm

And here's the fixed footer:
screen shot 2016-10-28 at 12 16 24 pm

### DIFF
--- a/templates/btax/results.html
+++ b/templates/btax/results.html
@@ -266,21 +266,14 @@
                         </tr>
                      `);
                      continue;
-                  }else{
-                    $table_body.append(`
-                      <tr class="group">
-                        <td colspan="7">
-                            <strong>${row.major_grouping}</strong> &emsp;
-                            Non-Corporate: ${row.summary_nc} &emsp;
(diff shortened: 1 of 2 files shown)


MattHJensen:
@PeterDSteinberg @jdebacker, in all situations like the one below, the second line should replace the first line, right? 

image

jdebacker:
Matt is right. 

Cc @peterdsteinberg

PeterDSteinberg:
ok will fix the double-summary-row issue.

PeterDSteinberg:
Last commit removes the double-summary row issue:

screen shot 2016-10-28 at 1 07 29 pm

jdebacker:
LGTM

PR #430, opened 2016-12-09
move logic to B-Tax for formatting its tables

### LINKED ISSUE #422: Cost of capital tables - "major industry" values inconsistent with subindustry results
@salimfurth points out:

>One thing that’s puzzling me is that a lot of the sector level results seem implausible as averages of their major industry components.
>  
> Considering “baseline” reported results only:
> ·         METR for typically-financed corporate manufacturing is 0.200, but the major industries in manufacturing have METR’s in the range [0.035, 0.199].
> ·         METR for debt-financed corporate manufacturing is 0.086, but the major industries are in [-0.083,0.082]
> o   Across all financing and corp/non-corp columns, the average looks like 100% weight has been assigned to “Chemical products” and some rounding errors or have been introduced.
> ·         METR’s for Information (all columns) are lower than the METR for any of the component major industries.
> ·         METR’s for Professional services (all columns) are lower than the METR for any of the component major industries.
> ·         METR’s for Health Care & Social Assistance (all columns) are higher than the METR for any of the component major industries.

I can confirm this behavior on the CCC web app.  However, when I run B-Tax master locally, the results look correct.

PeterDSteinberg:
Fixes #422 and will simplify the creation of B-Tax results.

### DIFF
--- a/webapp/apps/btax/helpers.py
+++ b/webapp/apps/btax/helpers.py
@@ -41,11 +41,7 @@ BTAX_ECON = ['btax_econ_nomint', 'btax_econ_inflat',]
 BTAX_VERSION_INFO = btax._version.get_versions()
 BTAX_VERSION = ".".join([BTAX_VERSION_INFO['version'], BTAX_VERSION_INFO['full-revisionid'][:6]])
 
-# Row labels, in order, including minor headings like "Durable goods"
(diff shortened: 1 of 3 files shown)


webapp/apps/btax/helpers.py
```diff
@@ -168,7 +164,7 @@ def hover_args_to_btax_depr():
     hover_notes['bonus_note'] = defaults['btax_depr_hover_exp']['notes']
     return hover_notes
 
-
+used = None
```
talumbau: Is this needed?

PeterDSteinberg: will remove - old debugging thing

webapp/apps/btax/views.py
```diff
@@ -323,10 +322,9 @@ def output_detail(request, pk):
 
     model = url.unique_inputs
     if model.tax_result:
-        output = url.unique_inputs.tax_result
+        tables = url.unique_inputs.tax_result[0]
```
talumbau: Is the new tax_result a list? Just wondering about the `[0]`.

PeterDSteinberg: The result returned by dropq (`tax_result`) is always a list, I think as long as the number of separate jobs, 1 in this case.

talumbau: hmm.. it looks to me like `tax_result` is always set as the return value of `dropq_get_results`, which is a dictionary, right?  For example, this line:

```
            model.tax_result = dropq_compute.dropq_get_results(normalize(job_ids))
```

webapp/apps/btax/helpers.py
```diff
@@ -215,101 +211,3 @@ def group_args_to_btax_depr(btax_default_params, asset_yr_str):
...
     return depr_argument_groups
 
-def btax_results_to_tables(results, first_budget_year):
-    """
-    Take various results from i.e. mY_dec, mX_bin, df_dec, etc
-    Return organized and labeled table results for display
-    """
-    asset_col_meta = dict(btax.parameters.DEFAULT_ASSET_COLS)
-    industry_col_meta = dict(btax.parameters.DEFAULT_INDUSTRY_COLS)
-    r = results[0]
```
PeterDSteinberg: @talumbau Commenting on the `tax_result[0]` discussion here: See the `r = results[0]` here.  This is how we were doing it (taking the 0th item of the return value from dropq)

PR #439, opened 2016-12-31
Only allow links to dynamic simulations on non-behavior run

### LINKED ISSUE #431: new features in json file
@martinholmer recently augmented the JSON reform file to include Growth, Behavior, and Consumption parameters in https://github.com/open-source-economics/Tax-Calculator/pull/1083

These changes break backwards compatibility for TB and TC, and some updates to TB are required before we can deploy any version of TC greater than 0.7.2 to TB. 

I described the core issue in https://github.com/open-source-economics/Tax-Calculator/pull/1083: 

> it won't be trivial to adopt on TaxBrain because of how TaxBrain guides users to static modeling first and then to dynamic modeling. Right now, partial eq and other dynamic runs come after the static results; with the new file upload feature, users will be able to jump directly to a partial equilibrium run without first generating static results. I don't think it will be hard to deal with the user interface and backend changes that will need to result from that, but it won't be trivial either.

To be more specific, the problem is that the new JSON format allows users to upload a file that will produce a partial equilibrium result rather than to a static result. The resulting page shouldn't have the same functionality as a static results page, and it also probably can't have the same functionality as the partial equilibrium result page.  

The reason that the page shouldn't have the same functionality as the static results page is that it won't make sense for the user to continue on to dynamic modeling after having specified behavioral parameters. 

The reason that the page probably can't have the same functionality as the partial equilibrium result page is that our existing PE page allows the user to return to the static results, but no static results were generated with the json file. 

I see two options for when the user includes nonzero behavioral parameters: 

1) Have the page take the user to what looks like the static results page, but when the user clicks "Link to Dynamic Simulations" a pop up will say "Your simulation is already a partial equilibrium dynamic simulation. Please resubmit without behavioral parameters if you would like to link to other dynamic modeling options." 

2) Have the page look like the partial equilibrium results page and either:
a) remove the "The microsimulation upon which this dynamic simulation was based can be found here"  or,
b) (fancier version) leave the "The microsimulation upon which this dynamic simulation was based can be found here" and when the user clicks "here", run that static calculation and generate the static results on the fly. 

@talumbau, it would be helpful to know which of these is easier to implement, and if there will be any other difficulties on the back end with incorporating the new json features from https://github.com/open-source-economics/Tax-Calculator/pull/1083

talumbau:
- A file-based reform could have behavioral parameters. If this is the
   case, you should not be able to launch a dynamic simulation.
 - Adds a modal pop-up if user clicks Link to Dyn. Sims on results page
   for file based reform that includes behavior parameters.
 - Behavior is unchanged if file based reform contains no behavioral
   components
 - Depends on change to worker node
 - Resolves #431

### DIFF
--- a/templates/taxbrain/results.html
+++ b/templates/taxbrain/results.html
@@ -60,7 +60,7 @@
         <a href="/dynamic/behavioral/edit/{{ unique_url.pk }}/?start_year={{ first_year }}" class="text-white btn btn-secondary">Edit Parameters</a>
         {% endif %}
         {% if is_micro %}
-        <a href="/dynamic/{{ unique_url.pk }}/?start_year={{ first_year }}" class="text-white btn btn-secondary">Link to Dynamic Simulations</a>
+        <button onclick="buttonAction()" class="text-white btn btn-secondary">Link to Dynamic Simulations</button>
         {% endif %}
       </div>
       {% if file_contents %}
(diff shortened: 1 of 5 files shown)


talumbau:
@MattHJensen here is what the modal pop-up looks like. Is the text ok?

screen shot 2016-12-31 at 12 11 12 pm

talumbau:
depends on:
https://github.com/OpenSourcePolicyCenter/deploy/pull/19
and
https://github.com/open-source-economics/Tax-Calculator/pull/1113

MattHJensen:
+1

PR #442, opened 2017-01-03
Change Default Year

### LINKED ISSUE #438: Switch TaxBrain default year to 2017
Right now the default year for TaxBrain simulations is 2016. Should it switch to 2017 and if so, when?

brittainhard:
Closes #438


brittainhard:
@MattHJensen @talumbau is there a need to change the tests to use 2017 as the start year?

talumbau:
confirmed that this works locally with `START_YEAR` environment variable. Also I like the addition of `/apps/constants.py`. Merging and then will update ospc.org.

PR #443, opened 2017-01-03
Remove JSON Endpoint

### LINKED ISSUE #355: remove JSON text field input page
the `/json` page has been removed in favor of `/file`. See #312, #353 for details.

brittainhard:
Closes #355


talumbau:
@brittainhard I did a global search for "START YEAR" (case insensitive). I think there are more places to change. It looks like there are hard coded "2016"s in there, but the desire was to make everything based on a global environment variable, as shown here:

https://github.com/OpenSourcePolicyCenter/webapp-public/blob/a37f4e50854d2b393e72e36405cd1c26ceae18bf/webapp/apps/taxbrain/helpers.py#L18

Let's change the default value to '2017' in the absence of any specified start year at that line.

brittainhard:
@talumbau this should be fixed in #442

talumbau:
Oops you are right. The above comment was fro #442.

talumbau:
ok this one looks good. merging.

PR #446, opened 2017-01-05
Update Install Steps in Readme

brittainhard:
Readme was good, only needed a small amount of changes, and a mention of the need for a script containing the environment files.


brittainhard:
This is ready for review. These commits should probably be squashed since they only apply to one file.

talumbau:
this is great. Exactly what was needed. +1. @brittainhard if you want to squash commits prior to merging that's fine. If no activity in a few days, I'll just merge.

cc @MattHJensen

PR #467, opened 2017-02-02
Remove column sorting (WIP)

brittainhard:


brittainhard:
Need to fix rendering problems with the industry table before this can be merged.

PR #475, opened 2017-02-16
Fix cell index ccc

PeterDSteinberg:
We were seeing Javascript errors in loading `/ccc` results page.  The errors related to index errors in datatables.js.  This PR ensures every row added to table has the same number of <tr><td> elements and thus avoids index error  in datatables.js


PeterDSteinberg:
cc @brittainhard

PR #483, opened 2017-02-23
Add Docs Page

### LINKED ISSUE #469: Add a separate "TaxBrain Documentation" page.
Rather than including so much documentation on the input pages, I think we could have a separate TaxBrain Documentation page. 

The content should be stored in a html file in webapp-public (for now) rather than in a flatblock, so it is easy for contributors to suggest changes via PRs. 

To begin with, we can keep everything in a single page / HTML file and link to relevant subsections with anchors. 

The page can use the same styling as our TaxBrain input pages. 

If no one has a better suggestion, url could be www.ospc.org/taxbrain/docs

brittainhard:
Closes #469


MattHJensen:
@brittainhard, let's not add a link to the header bar just yet: 
image

MattHJensen:
This looks great otherwise.

PR #484, opened 2017-02-24
Fix Edit Parameters Rendering

### LINKED ISSUE #459: Edit parameters page only shows font color change
Here is what an input field looks like to begin with. 

image

When I edit a parameter on the TaxBrain input page, the color of the font darkens, the input field turns green, and the Default value appears under the input field. This is great. 

image

However, when I run TaxBrain and then click edit parameters, only the font is darker:

image

It would be better if the edit parameters page also included the input field color changes and the default values appearing under the input page (i.e, it should look exactly like the second image from this post rather than the third). 

This issue applies to both TaxBrain and CCC.

brittainhard:
Closes #459


MattHJensen:
+1

PR #485, opened 2017-02-24
Remove edited class when reset button is pressed.

### LINKED ISSUE #420: Reset Parameter button should entirely reset input page.
On Chrome, after the "Reset Parameters" button is activated, the parameters that were used and then reset are left highlighted and displaying the default value subscript.

image

I think the button should return the page to its original form. One way to do that would have be to simply have it link to www.ospc.org/taxbrain. 

(using the browser's refresh capability won't work, as our original motivation for adding the reset parameter button is that in firefox, the refresh button does not clear out existing parameters (see #379). Reloading www.ospc.org/taxbrain works fine.)

brittainhard:
Closes #420

### DIFF
--- a/static/js/taxbrain.js
+++ b/static/js/taxbrain.js
@@ -99,7 +99,11 @@ $('.form-group > input.form-control').blur(function() {
   }
 });
 
-
+$("#tax-submit[type='reset']").click(function(){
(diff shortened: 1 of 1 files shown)


MattHJensen:
+1

PR #487, opened 2017-02-24
Embed Google Form at /hello

### LINKED ISSUE #471: Embed google form at ospc.org/hello
We have a new google form for feedback (see #406). Could we embed this at www.ospc.org/support or ospc.org/help (whichever others thing is better? I will work with @andersonfrailey to generalize the form to be applicable to the whole organization rather than just TaxBrain, but that needn't block this issue.

brittainhard:
Closes #471

### DIFF
--- a/templates/pages/hello.html
+++ b/templates/pages/hello.html
@@ -5,53 +5,7 @@
 {% load flatblocks %}
 
 {% block content %}
-<div class="text-center">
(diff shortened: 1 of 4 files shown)


MattHJensen:
@brittainhard, let's not add a link to the header bar just yet: 
image

MattHJensen:
Looks great otherwise.

PR #491, opened 2017-02-27
Change flatblock header names.

brittainhard:
cc @MattHJensen


templates/dynamic/results.html
```diff
@@ -41,7 +41,7 @@ <h2 class="sr-only">TaxBrain</h2>
 
     <div class="result-header">
       <div class="result-header-control">
-          <h1>{% flatblock "dynamic_results_header" %} {{ tables.result_years.0 }} - {{ tables.result_years | last }} </h1>
+          <h1>{% flatblock "taxbrain_pe_results_header" %} {{ tables.result_years.0 }} - {{ tables.result_years | last }} </h1>
```
MattHJensen: @brittainhard, this isn't the right place for the change. This is the [OG-USA model](http://www.ospc.org/dynamic/ogusa/5000/?start_year=2017). The issue is instead w the [partial equilibrium model](http://www.ospc.org/dynamic/behavioral/5000/?start_year=2017), labeled "behavioral". 

The problem is that right now the partial equilibrium model uses the taxbrain_results_header flatblock (after this PR, `taxbrain_static_results_header`). Instead, the partial equilibrium model needs a new flatblock, which is where `taxbrain_pe_results_header` should go. 

The `dynamic_results_header` is certainly confusingly named, but we can leave it alone for now.

brittainhard: Ahh, that makes sense. So the header name change in `taxbrain/results.html` is correct?

MattHJensen: > So the header name change in taxbrain/results.html is correct?

Yep.

MattHJensen:
Looks good on the test app. +1

PR #509, opened 2017-03-13
Travis env var fix

PeterDSteinberg:
Add an environment variable to .travis.yml to avoid keyerror in CI.


PeterDSteinberg:
@MattHJensen @brittainhard We should merge this into master then update the other PRs so /ccc tests don't fail in CI.

MattHJensen:
+1

PR #514, opened 2017-03-18
avg -> total in tables. 

### LINKED ISSUE #513: Mislabeled TaxBrain output table
When arriving at the static analysis results page on TaxBrain, I ask for a diagnostic table for a few variables calculated by income deciles.  What comes up seems to have a misleading title.  Here is an example of what I mean:
```
USER PLAN TAX VARS, WEIGHTED AVG PER EXPANDED INCOME DECILE (2020)
Income     Returns        AGI    IITliab     PTliab
Decile   Thousands   Billions   Billions   Billions
0-10%       17,780     -120.5       -2.5        7.0
10-20%      17,781      116.4      -13.6       16.1
etc
```
Given that the dollar amounts in the table are clearly WEIGHTED TOTALS for the decile, shouldn't the title be changed from WEIGHTED AVG PER to WEIGHTED TOTAL PER?

Or if the intention is to actually show WEIGHTED AVG amounts, then leave the title unchanged and divide each dollar entry by the number of returns on the table line (making sure to change "Billions" to "Thousands" in each of those dollar columns).

I don't have a preference for which numbers are shown, but do think it would leave a better impression with the public if the tables where titled correctly.

@MattHJensen @PeterDSteinberg @brittainhard

MattHJensen:
fixes #513

### DIFF
--- a/webapp/apps/taxbrain/helpers.py
+++ b/webapp/apps/taxbrain/helpers.py
@@ -225,13 +225,13 @@ TAXCALC_RESULTS_DEC_ROW_KEY_LABELS = {
     'all':'All'
 }
 TAXCALC_RESULTS_TABLE_LABELS = {
-    'mX_dec': 'Base plan tax vars, weighted avg per expanded income decile',
-    'mY_dec': 'User plan tax vars, weighted avg per expanded income decile',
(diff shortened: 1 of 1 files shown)


martinholmer:
Pull request #514 looks like it will fix issue #513, but it needs to be merged to actually get the fix.

PR #526, opened 2017-04-06
Move btax to conda-requirements.

brittainhard:


PeterDSteinberg:
How is ogusa installed now?  We should also fix OG-USA install in this PR.

brittainhard:
Not sure. I'll take a look.

brittainhard:
@PeterDSteinberg i dont see it anywhere that ogusa is installed in webapp

PeterDSteinberg:
ok - that's right (regarding OG-USA).  I think we can merge this PR, then.

PR #551, opened 2017-05-08
Form UI Additions

### LINKED ISSUE #417: Group TB (and CCC) inputs into an accordion menu
Over at https://github.com/open-source-economics/Tax-Calculator/issues/1074#issuecomment-261808528, @feenberg said:

> Perhaps the TB web page could have expandable sections, where you could select a section to expose the parameters available to set, and perhaps subsections that could be expanded recursively.

This seems like a good idea to me too. 

@talumbau and @brendancol, do you have any feedback before I add this to a milestone? Is there a relatively simple way to do this?

### LINKED ISSUE #504: _ALD_InvInc_ec_base_RyanBrady should be checkbox instead of 1/0
As a rule, we can use checkboxes when parameters are boolean in current_law_policy.json. 

image

@brittainhard

### LINKED ISSUE #520: NIIT parameter is Boolean
The `_NIIT_PT_taxed` parameter is Boolean in Tax-Calculator, while that is not the case in TaxBrain, as indicated in the screen shot. This seems a bug to me.

cc @MattHJensen @brittainhard 

screen shot 2017-04-04 at 1 37 57 pm

### LINKED ISSUE #554: Remove "looks okay" button and down arrow.
Let's get rid of these: 

image

They look particularly goofy w the accordion:

image

### LINKED ISSUE #555: Default values are no longer displayed after editing parameters
image

No default is displayed.

brittainhard:
Closes #520 
Closes #504 
Closes #417 
Closes #554 
Closes #555


brittainhard:
@MattHJensen if its okay with you im going to merge this and push it to master when we get the taxcalc update working.

MattHJensen:
@brittainhard, are all of these on the test app now? What's the reason for waiting for the taxcalc update?

brittainhard:
@MattHJensen this branch is up on the test app now. I think we can update the main app without issues.

MattHJensen:
@brittainhard, it looks like taxcalc 0.8.4 is on the test app. I thought we couldn't put 0.8.4 up because of the dropq issue?

brittainhard:
@MattHJensen That's correct. We still have some issues that we are working through with taxcalc. The UI elements in this PR are separate from taxcalc, so there should be no problem.

MattHJensen:
I see. I just tested this PR. Looks good. When we deploy to ospc.org we should use 0.8.3 for both front end and back end. Will deploying to ospc.org require downtime?

brittainhard:
@MattHJensen  When we update the front-end it will automatically start using taxcalc 0.8.4. The heroku build is set to to grab the latest packages. As far as I know, the backend will still be at 0.8.2. I can't access the nodes right now, but i'll double-check with Peter.

brittainhard:
Deploying and updating to 0.8.3 will take some time.

brittainhard:
@MattHJensen just checked with @PeterDSteinberg and he said that the nodes are at taxcalc 0.8.3.

MattHJensen:
> just checked with @PeterDSteinberg and he said that the nodes are at taxcalc 0.8.3.

That's good. 

Could you change something to make the heroku build grab a package you specify? And then I think it would be helpful to have a test that ensures that the nodes and front end are using the same package.

brittainhard:
@MattHJensen the only interface for updating heroku is git. If we wanted to modify the version it uses, we would have to modify the buildpack itself and hard-code it (buildpack is at https://github.com/OpenSourcePolicyCenter/conda-buildpack).

I don't think the UI update will cause errors even with the version being ahead of the nodes. Most of the changes were JavaScript related, and the changes that were made to the django part of the app should not cause errors -- they were database field updates to accommodate the checkbox changes.

MattHJensen:
@brittainhard, hardcoding sounds like the right solution. Going forward I think it will be important to be able to keep the nodes and front end aligned, even if it isn't necessarily an issue w this particular release.

On May 18, 2017, at 11:24 AM, Brittain Hard <notifications@github.com<mailto:notifications@github.com>> wrote:


@MattHJensen<https://github.com/matthjensen> the only interface for updating heroku is git. If we wanted to modify the version it uses, we would have to modify the buildpack itself and hard-code it (buildpack is at https://github.com/OpenSourcePolicyCenter/conda-buildpack).

I don't think the UI update will cause errors even with the version being ahead of the nodes. Most of the changes were JavaScript related, and the changes that were made to the django part of the app should not cause errors -- they were database field updates to accommodate the checkbox changes.

Issue #534, opened 2017-04-13
TaxBrain "Build Table" enhancements


martinholmer:
The "Build Table" feature at the bottom of the TaxBrain static results page is very nice.  I've used it off and on over the past year mostly by building a table and viewing it and maybe making a note of a few numbers that I'm viewing.

I've just started exploring the other options of what to do with a built table: Print, Copy and CSV.
I don't have Excel on my computer so I didn't try the fourth option.

Here are a few things I noticed that could be improved to make the user experience better.

(1) When using Print, the table looks good on the page, but a second blank page gets printed out even though there is plenty of whitespace at the bottom of the first page.  It's sort of a hassle to have to remember to switch the print option from All to Page 1, and even if you know to do that, the one printed page says it is 1/2, which is confusing.  Someone might think they were missing some output.

(2) When using Copy, everything works fine except the title of the built table is absent.  This is a major shortcoming.  The built table's title is included in the Print output, so I don't know why it wouldn't be included in the Copy output.

(3) When using CSV, the downloaded file loads up into my spreadsheet (OpenOffice) just fine except again the critically important title of the built table is missing.

I'm a novice user of these TaxBrain capabilities, so I could have made a mistake.  But if not, fixing these things would leave users would a much better impression of TaxBrain.

@MattHJensen @PeterDSteinberg @brittainhard

martinholmer:
Another enhancement needs to be added to the list in TaxBrain issue #534.

It has already been pointed out that the title of the built table is not included in the downloaded CSV file or Copy.  But there is another problem: the name of the downloaded file is sort of laughable and completely useless.  Here is what I got in my `~/Download` directory after building four different distributional tables and clicking on the CSV button:
```
  989 Apr 14 17:09   Welcome to TaxBrain  TaxBrain (1).csv
  976 Apr 14 17:10   Welcome to TaxBrain  TaxBrain (2).csv
  961 Apr 14 17:12   Welcome to TaxBrain  TaxBrain (3).csv
  640 Apr 14 17:07   Welcome to TaxBrain  TaxBrain.csv
```
First of all, the embedded spaces cause failures with standard CLI tools.
Here is what I get on my Mac:
```
iMac2:tax-calculator mrh$ ll ~/Downloads/Welc*
ls: (1).csv: No such file or directory
ls: (2).csv: No such file or directory
ls: (3).csv: No such file or directory
ls: /Users/mrh/Downloads/Welcome: No such file or directory
[...]

PR #558, opened 2017-06-07
Modify Naming Conventions for CSV Output File and Excel Output File

GoFroggyRun:
This PR tries to partially resolve concerns addressed in #534. 

In particular, @MattHJensen suggested naming scheme for csv output files and excel output files in [534 comment](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/534#issuecomment-296374906). This PR mostly follows such scheme, expect substituting suggested keywords "payroll-income" and 'income" with "combined" and "individual" respectively so that the file name corresponds to the title of the table. 

Based on local tests, the output file names read: 

screen shot 2017-06-09 at 1 17 26 pm

Note that 127 is my run number, which will be changing according to the url. Please also note that the some buttons are not working for total liabilities tables, and currently the only output file for that table is in the format `xxx_liabilities_change`.

@MattHJensen Could you review? Any comments, concerns or remarks are welcomed.


MattHJensen:
@GoFroggyRun, is this ready for review? 

cc @brittainhard

GoFroggyRun:
Yes. All I have are included in this PR.


On Jun 13, 2017, at 5:11 PM, Matt Jensen <notifications@github.com<mailto:notifications@github.com>> wrote:


@GoFroggyRun<https://github.com/gofroggyrun>, is this ready for review?

cc @brittainhard<https://github.com/brittainhard>

requirements.txt
```diff
@@ -8,7 +8,6 @@ static3==0.5.1
 wsgiref==0.1.2
 -e git+https://github.com/funkybob/django-flatblocks.git#egg=django-flatblocks
 sendgrid-django
-psycopg2==2.5.4
```
brittainhard: Whats the reasoning behind removing this dependency?

GoFroggyRun: @brittainhard: 
When I was trying to set up everything locally with @PeterDSteinberg, some dependency error message regarding `psycopg2` pops out. Removing this dependency seems to resolve such issue.

brittainhard: What's the dependency issue? Can you get a traceback?

brittainhard: I'm mainly concerned that this will cause issues on the heroku deployment, since it uses postgres as a database.

GoFroggyRun: @brittainhard: 
I wasn't able to locate the traceback since it was a while ago. Currently my conda list shows that `psycopg2` is on 2.7.1. Would this be a problem for heroku deployment?

talumbau:
Yes, it very likely will. There must be something to setup the Python
environment to talk to Postgres.

requirements.txt
GoFroggyRun: In other words, would it be a problem if I change the line 
```
-psycopg2==2.5.4
```
to 
```
-psycopg2
```
that removes the version requirement?

martinholmer: How about this:
```
psycopg2>=2.5.4
```
Won't that satisfy everybody?

brittainhard: @GoFroggyRun if the main issue that you are having issues using `postgres` locally? If thats the case, there is the option of using SQLite for local development. It's what i use, since I have to alter the database frequently.

Is the problem that your `psycopg2` version is causing problems, so you had to upgrade it? In that case @martinholmer that's a reasonable fix.

GoFroggyRun:
Thanks @martinholmer and @brittainhard for your advices. It's the version of `psycopg2` that bothers me, and I adopted @martinholmer's suggestion to fix that. Please take another look.

brittainhard:
Looks good to me. +1

PR #564, opened 2017-06-19
[WIP] Limit PDF/ Print Option to Produce Only One Page

GoFroggyRun:
Per discussion in [#534](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/534), the `Print` button on TaxBrain output page currently produces two pages. See the following screenshot for detail:

pdf_one_page

I tried different ways to limit the number of pages to print via `DataTable.ext.buttons.pdfHtml5` and `DataTable.ext.buttons.print` functions (details could be found within the commit), but they do not seem to help. 

@brittainhard, could you advice what's the best approach to deal with this issue?

### DIFF
--- a/static/js/vendor/DataTables/datatables.js
+++ b/static/js/vendor/DataTables/datatables.js
@@ -27845,7 +27845,8 @@ DataTable.ext.buttons.pdfHtml5 = {
 
 		var doc = {
 			pageSize: config.pageSize,
-			pageOrientation: config.orientation,
+            maxPagesNumber: 0, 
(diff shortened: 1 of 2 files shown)


brittainhard:
Print output looks great, +1

brittainhard:
@GoFroggyRun looks like we have more merge conflicts. In the future, I think it's best not to modify the vendor files themselves, but use a module separate script that changes the attributes of the original objects. I didn't think about this with the first PRs you made, but I think this code is fine for now.

PR #567, opened 2017-06-20
Fixes for CPI Calculation (WIP)

brittainhard:


brittainhard:
Merging this. The travis build is restricted to using the main channel version of taxcalc, which is still at 0.8.4. I tested it locally, and the build will be fixed as soon as travis can use the latest taxcalc.

PR #568, opened 2017-06-20
Change Title Terminology for Tables

GoFroggyRun:
As discussed in [#534](https://github.com/OpenSourcePolicyCenter/webapp-public/issues/534#issuecomment-296374906), the terminology for the diagnostic tables and the difference tables are different. 

The is old title name:
screen shot 2017-06-20 at 5 00 22 pm

This is the title name proposed in this PR:
screen shot 2017-06-20 at 5 00 11 pm

All "Per" are changed to "By" so that the diagnostic tables and the difference tables have consistent title names. 

@brittainhard @MattHJensen Could you review?


brittainhard:
Looks straightforward, +1.

Issue #578, opened 2017-07-02
Regression in TaxBrain file upload logic


martinholmer:
The following message was received from TaxBrain when uploading reform and assumption files that have been successfully uploaded for many months (see the history of the `tax-calculator/file-upload-tests/README.md` file).  First the error and then the uploaded files.

![image](https://user-images.githubusercontent.com/12170745/27773585-f00ebb44-5f4a-11e7-955a-3eb84660ea00.png)

The uploaded files are `r1.json` and `a0.json` in the `tax-calculator/file-upload-tests` directory.  The above error is generated even though the START YEAR parameter on the TaxBrain file-upload page is set to 2013, which means I was asking for results for the ten years from 2013 through 2022.

It would seem the problem is with TaxBrain because using the taxcalc 0.9.0 package on my local computer runs just fine:
```
iMac2:file-upload-tests mrh$ conda list taxcalc
# packages in environment at /Users/mrh/anaconda:
#
taxcalc                   0.9.0                    py27_0    local

iMac2:file-upload-tests mrh$ tc puf.csv 2022 --reform r1.json --assump a0.json 
You loaded data for 2009.
Tax-Calculator startup automatically extrapolated your data to 2022.

iMac2:file-upload-tests mrh$ ls -l puf-22-r1-a0.csv 
-rw-r--r--  1 mrh  staff  8802722 Jul  2 17:24 puf-22-r1-a0.csv
iMac2:file-upload-tests mrh$ 
```
And the full set of local tests runs just fine:
```
iMac2:file-upload-tests mrh$ ./cli-results.sh 
STARTING WITH CLI-RESULTS : Sun Jul  2 17:08:08 EDT 2017
outfile,inctax,paytax= puf-22-r1-a0.csv 1823.04  1466.41
outfile,inctax,paytax= puf-22-r1-a1.csv 1775.99  1439.72
outfile,inctax,paytax= puf-22-r1-a2.csv 1800.40  1454.31
FINISHED WITH CLI-RESULTS : Sun Jul  2 17:09:22 EDT 2017
```

@MattHJensen @PeterDSteinberg @brittainhard @jbcrail

brittainhard:
@martinholmer will take a look at this sometime today.

martinholmer:
@brittainhard said:
> will take a look at this [#578] sometime today.

Thanks.  Now that we have main channel version of taxcalc 0.9.0 package, I can confirm that using that package on my computer produces results without error.

[...]

Issue #557, opened 2017-05-30
Remove Year Dropdown from File Upload Page


brittainhard:
Currently has a dropdown for adding a start year, but doesn't seem to
change the outcome of the data sent forward, since the data provided
contains the year directly. Should probably remove this UI element.

Of course, there may be some code that allows for a differently formatted
file -- indeed, there is really no indication that I can see where
a format has been specified.

If there were format specifications, I think they would be on the server,
but from looking at the code, it doesn't look like there is anything there.

MattHJensen:
@brittainhard the year drop down specifies the start of the 10-year budget window. The reform files themselves specify the year that various prvisions of the reform are implemented. 

Here is an example of a reform where the year drop down was set to 2013. http://www.ospc.org/taxbrain/13557/

martinholmer:
There has been no discussion of issue #557 over the past four weeks.
So, I'm assuming this issue can be closed.

Issue #543, opened 2017-05-02
Simplify deploy / webapp-public interaction 


PeterDSteinberg:
We have had bugs related to the interaction of this repo and the private deploy repo in API functions, such as passing jsons between the front end Django app (now webapp-public) and the back end services (celery - celery_tasks.py, flask - flask_server.py).  Goals of this issue are to plan:

 * [x] Move most of deploy assets to this repo at a top level subfolder called `deploy`: python files mostly to subpackage, bash scripts case by case review based on how they are used.  
 * [x] Change 2 existing API styles (file based reforms and UI reforms) in webapp / deploy have one API, so that file reforms or UI reforms go to same end point.  Try to de-deduplicate code.
 * [x] Improve backwards compatibility of models in one API refactor
 * [ ] Make a mocked system test harness so that webapp-public can prepare arguments and jsons as though it is going to submit them to the flask / celery system of deploy but instead calls Python function

Make one PR soon that moves the deploy repo to webapp-public and possibly some minor other changes while trying not affect existing usage of webapp.  Then do >=1 PRs to clean up and address the bullets above.

Thoughts?  @MattHJensen @brittainhard

MattHJensen:
+1

martinholmer:
What is the status of issue #543?  Haven't some of the pull request merges of the past few weeks made progress on items listed in the first comment in #543?

@brittainhard @PeterDSteinberg @jbcrail @MattHJensen

PeterDSteinberg:
Hi @MattHJensen  - Regarding the last 2 checkboxes above:
 * Has backwards compatibility been okay over the last month from your perspective?  Any models you have run in the last month or 6 weeks that are no longer displaying results correctly?  Let's discuss backwards compatibility again when we plan the OG-USA work, such as the meeting on Friday of this week.  Backwards compatibility is less challenged by changes to the input forms to models, but when we do major changes to the results pages, such as addition of new tables or figures or interactivity, we should make backwards compatibility a check list item for the PR(s).  
* Bullet 4 ` Make a mocked system test harness so that webapp-public can prepare arguments and jsons as though it is going to submit them to the flask / celery system of deploy but instead calls Python function` is not addressed yet.  I'll make a separate issue for it - I think it will make the tests more exhaustive to have Django CI tests or local Django usage be able to dump a json to file or one's browser without calling out to flask.

PeterDSteinberg:
#580 covers the final bullet point.

PR #593, opened 2017-07-27
File Upload Fixes

brittainhard:
@PeterDSteinberg @martinholmer @MattHJensen @hdoupe

I was unable to add a unit test that could help us solve the problem with this file, mainly because it would require saving `puf.csv.gz`. I was, however, able to make a small script that duplicates the behavior that is causing the problem.

```
import json
import celery

import pandas as pd
import taxcalc
from deploy.taxbrain_server.celery_tasks import dropq_task

behavior_params = {u'growdiff_response': {}, u'consumption': {}, u'behavior': {}, u'growdiff_baseline': {}}

user_mods = {u'2019': {u'_SS_Earnings_c': [500000], u'_ALD_InvInc_ec_rt': [0.2]}, u'2020': {u'_SS_Earnings_c': [600000], u'_II_em_cpi': True}, u'2018': {u'_SS_Earnings_c': [400000], u'_II_em_cpi': False, u'_II_em': [8000]}}

puf = pd.read_csv("puf.csv.gz", compression='gzip')

results = dropq_task(0, user_mods, None, behavior_params, puf)
```

You have to run it from the repository root, and `puf.csv.gz` has to also be in the root. This way, you can test out the function and make changes without having to use celery.

Let me know if it helps.

EDIT: You can cause the error by increasing the `year_n` argument, the first argument, to something like 9 or 10.

### DIFF
--- a/deploy/taxbrain_server/celery_tasks.py
+++ b/deploy/taxbrain_server/celery_tasks.py
@@ -60,18 +60,14 @@ def dropq_task(year, user_mods, first_budget_year, beh_params, tax_data):
     # index indicates whether the reform dictionary was non-empty
     # The four reform dictionaries from file-based reforms are:
     # policy, behavior, growth, consumption (in that order)
-    reform_style = [True]
-    if first_budget_year is not None:
-        first_year = int(first_budget_year)
-    else:
-        first_year = int(user_mods.keys()[0])
-
+    first_budget_year = int(first_budget_year)
     user_mods = convert_int_key(user_mods)
-    print('first_year', first_year)
-    if user_mods.get(first_year):
-        for key in set(user_mods[first_year]):
-            if key.startswith('_BE_'):
-                user_mods[first_year].pop(key)
+    print('first_year', first_budget_year)
+    for reform_year in user_mods.keys():
+        if user_mods.get(reform_year):
+            for key in set(user_mods[reform_year]):
+                if key.startswith('_BE_'):
+                    user_mods[reform_year].pop(key)
         user_reform = {"policy": user_mods}
     print('user_reform', user_reform, user_mods)
     reform_style = [True if x else False for x in user_reform]
@@ -84,7 +80,7 @@ def dropq_task(year, user_mods, first_budget_year, beh_params, tax_data):
     for key in EXPECTED_KEYS:
         if key not in user_reform:
             user_reform[key] = {}
-    kw = dict(year_n=year, start_year=first_year,
+    kw = dict(year_n=year, start_year=first_budget_year,
               taxrec_df=tax_data, user_mods=user_reform)
     print('keywords to dropq', {k: v for k, v in kw.items()
                                 if k not in ('taxrec_df',)})
--- a/webapp/apps/taxbrain/compute.py
+++ b/webapp/apps/taxbrain/compute.py
@@ -118,7 +118,8 @@ class DropqCompute(object):
         hostnames = workers[dropq_worker_offset: dropq_worker_offset + num_years]
         print "hostnames: ", hostnames
         num_hosts = len(hostnames)
-        data['user_mods'] = json.dumps(user_mods)
+        data["user_mods"] = json.dumps(user_mods)
+        data["first_budget_year"] = str(first_budget_year)
         if additional_data:
             if "behavior" in additional_data.keys():
                 data["behavior_params"] = json.dumps(additional_data)
--- a/webapp/apps/taxbrain/compute.py
+++ b/webapp/apps/taxbrain/compute.py
@@ -118,6 +118,7 @@ class DropqCompute(object):
         hostnames = workers[dropq_worker_offset: dropq_worker_offset + num_years]
         print "hostnames: ", hostnames
         num_hosts = len(hostnames)
+        data['first_budget_year'] = json.dumps(first_budget_year)
         data['user_mods'] = json.dumps(user_mods)
         if additional_data:
             if "behavior" in additional_data.keys():
--- a/webapp/apps/taxbrain/compute.py
+++ b/webapp/apps/taxbrain/compute.py
@@ -118,7 +118,6 @@ class DropqCompute(object):
         hostnames = workers[dropq_worker_offset: dropq_worker_offset + num_years]
         print "hostnames: ", hostnames
         num_hosts = len(hostnames)
-        data['first_budget_year'] = json.dumps(first_budget_year)
         data['user_mods'] = json.dumps(user_mods)
         if additional_data:
             if "behavior" in additional_data.keys():
--- a/webapp/apps/taxbrain/tests/test_compute.py
+++ b/webapp/apps/taxbrain/tests/test_compute.py
@@ -18,6 +18,7 @@ SUBMIT_DROPQ_DATA = {
     "url_template": 'http://{hn}/dropq_start_job',
     "additional_data": {'growdiff_response': {}, 'consumption': {}, 'behavior': {}, 'growdiff_baseline': {}},
     "is_file": True
+}
 
 
 class MockComputeTests(MockCompute):
--- a/webapp/apps/taxbrain/tests/test_compute.py
+++ b/webapp/apps/taxbrain/tests/test_compute.py
@@ -1,0 +1,47 @@
+import mock
+import json
+
+from django.test import TestCase
+from django.test import Client
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.test.client import RequestFactory
+
+from .. import compute
+from ..compute import DropqCompute, MockCompute
+from ...test_assets import *
+import requests_mock
+
+
+SUBMIT_DROPQ_DATA = {
+    "mods": {2018: {u'_SS_Earnings_c': [400000], u'_II_em_cpi': False, u'_II_em': [8000]}, 2019: {u'_SS_Earnings_c': [500000], u'_ALD_InvInc_ec_rt': [0.2]}, 2020: {u'_SS_Earnings_c': [600000], u'_II_em_cpi': True}},
+    "first_budget_year": 2013,
+    "url_template": 'http://{hn}/dropq_start_job',
+    "additional_data": {'growdiff_response': {}, 'consumption': {}, 'behavior': {}, 'growdiff_baseline': {}},
+    "is_file": True
+
+
+class MockComputeTests(MockCompute):
+
+    def remote_submit_job(self, theurl, data, timeout):
+        assert "first_budget_year" in data.keys()
+        return super(MockComputeTests, self).remote_submit_job(theurl, data, timeout)
+
+
+class ComputeTests(TestCase):
+    ''' Test the views of this app. '''
+
+    def setUp(self):
+        # Every test needs a client.
+        self.dropq_compute = MockComputeTests()
+        # self.dropq_compute.remote_submit_job = MockCompute().remote_submit_job
+        self.client = Client()
+
+    def test_submit_calculation(self):
+        self.dropq_compute.submit_calculation(
+            SUBMIT_DROPQ_DATA["mods"],
+            SUBMIT_DROPQ_DATA["first_budget_year"],
+            SUBMIT_DROPQ_DATA["url_template"],
+            num_years=1,
+            additional_data=SUBMIT_DROPQ_DATA["additional_data"],
+            pack_up_user_mods=False
+        )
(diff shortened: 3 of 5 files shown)


brittainhard:
Also, I think its a good idea if we move the discussion here from now on.

MattHJensen:
@hdoupe, could you explore whether the [cps.csv.gz](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/cps.csv.gz) file in tax-calculator could be used to write failing tests for the bugs described in #578?

brittainhard:
@martinholmer I think maybe I am missing some understanding here.

So when I use the regular taxbrain form (not file input) to do calculations, and I set my start year, it calculates 9 years from that start date. So, if my start date is 2015, it calculates up to 2024.

When parsing the `r1.json` file, the earliest parameter date is 2018. That means that when the data is sent out, it does not contain parameters for any year before 2018. As is, the application assumes the start date is 2018 and it throws an error. With file uploads, the drop-down menu that allows you to select the year does not have any outcome on the data you send, because its is relying only on the data in the file. In this case, the year selection drop-down is useless.

So what does the ideal outcome look for you? If we use that sample file, do you want calculations from 2018 until 2026? That is a straightforward fix.

Please let me know.

martinholmer:
@brittainhard said:
> So when I use the regular taxbrain form (not file input) to do calculations, and I set my start year, it calculates 9 years from that start date. So, if my start date is 2015, it calculates up to 2024.

Yes, exactly.

> When parsing the r1.json file, the earliest parameter date is 2018. That means that when the data is sent out, it does not contain parameters for any year before 2018. As is, the application assumes the start date is 2018 and it throws an error. With file uploads, the drop-down menu that allows you to select the year does not have any outcome on the data you send, because its is relying only on the data in the file. In this case, the year selection drop-down is useless.

That's one of the problems.  The drop-down menu of start_year values should work **exactly the same** on the file upload page as it does on the regular TaxBrain form.  The start_year is the first of the ten budget years, for which we show results.  There is no problem If the tax reform is not implemented until several years into the ten budget year window; the pre-reform years will just show no change in tax revenue.  Perhaps some of the confusion is caused by the fact that the way to specify on the regular TaxBrain form a reform that is first implemented after the start_year is not very well documented.  If start_year is 2017 (the default) and you want to make all earnings taxable under the payroll tax beginning in 2020, you would enter the following in the appropriate box:
```
*,*,*,9e99
```
The asterisk symbol means use the value under current law policy and 9e99 is a very large number for the maximum social security taxable earnings parameter.  I recommend that you try this to see how it works from the regular TaxBrain form.

> So what does the ideal outcome look for you? If we use that sample file, do you want calculations from 2018 until 2026?

No, we always want ten budget years.  What we need is exactly the same as on the regular TaxBrain form: start_year specified by the pull-down menu (which has 2017 as the maximum allowable value) and nine more years after the start_year.

Does this make sense?  If not, ask another question before doing any more coding.

hdoupe:
@MattHJensen 
could you explore whether the cps.csv.gz file in tax-calculator could be used to write failing tests for the bugs described in #578?

Sure thing.  I'll look into it.

brittainhard:
@martinholmer 

> That's one of the problems. The drop-down menu of start_year values should work exactly the same on the file upload page as it does on the regular TaxBrain form. The start_year is the first of the ten budget years, for which we show results. There is no problem If the tax reform is not implemented until several years into the ten budget year window; the pre-reform years will just show no change in tax revenue.

I think this highlights a fundamental problem with the start-year dropdown.

First, if you want to display an entire 10 year window and your parameters start at a year later than the year specified in the dropdown, you're going to end up with columns in the table that contain no useful information.

Second, suppose you specified your start year as 2013 and your first parameters in the reform file started at 2023. You would get a table containing only columns without data, and your parameters would be completely ignored -- the table would only show data up to the year 2022. Moreover, in this case, if we were to make this change to show all 10 years even if there is no data, that can be considered a regression in itself.

From a user experience standpoint, this is a problem. Having a table that contains useless information or no information at all does not help the user understand the effects of their reform, and it can confuse the user about how these file-based reforms work. If they are receiving useless information, they might think there is something wrong with their parameters file, when that is not at all the case.

The solution I propose is this: include some code that will limit the number of years to be calculated if the start date is after 2017 (this will handle the error), and remove the start year drop-down entirely. This will mean the start year will be the earliest year in the reform file. 

With that in mind, I have reopened issue #557. I think this is the best way to handle this problem.

@MattHJensen @PeterDSteinberg @hdoupe let me know what you think.

MattHJensen:
@brittainhard 

The start_year drop down should specify the beginning of the budget window. 

If the user makes no parameter changes within the ten-year budget window, then all revenue changes should be zero (not missing), and the current_law and reform revenues should be the same (again, not missing). 

If the user's first parameter change is for the 6th of the budget window (as defined by the start year dropdown), then the revenue changes for the first five years should be 0 and non-zero in year 6. 

All of this is the same as what @martinholmer describes, https://github.com/OpenSourcePolicyCenter/webapp-public/pull/593#issuecomment-318549498.


>Second, suppose you specified your start year as 2013 and your first parameters in the reform file started at 2023. You would get a table containing only columns without data, and your parameters would be completely ignored -- the table would only show data up to the year 2022. Moreover, in this case, if we were to make this change to show all 10 years even if there is no data, that can be considered a regression in itself.

We could add a warning that "your first parameter change happens outside of the budget window."

MattHJensen:
Here is a test that you could write: 

-Run a reform file where the start year is 2017 and the only reform provision is to set the social security payroll tax rate to 10% is 2019. 

- Run a GUI reform where the the start year is 2017 and the only parameter reform is to set the social security payroll tax rate to: `*, *, 0.10`

- Assert that the two outputs are the same.

brittainhard:
@MattHJensen I can make that fix. All I need to do is supply dummy information prepended to the reform data depending on what start year is selected. After that, we will need some tests to handle the case I mentioned before, with the start year of 2013. This will require some change in the `file_input` view, and another small change in `celery_tasks.py`.

MattHJensen:
>@MattHJensen I can make that fix. 

Thanks. 

>All I need to do is supply dummy information prepended to the reform data depending on what start year is selected.

Conforming to the goal of a unified API for reform files and GUI based inputs, I'd suggest following the same technique that is used for the asterisk (*) operator. 

> After that, we will need some tests to handle the case I mentioned before, with the start year of 2013. This will require some change in the file_input view, and another small change in celery_tasks.py.

I'd suggest handling this as a separate issue and supplying a warning rather than an error. There are legitimate reasons why a user would set the start year at 2013 and use a reform file with provisions starting in 2025. For example, s/he might be running two reforms with the same reform file, one with the start year at 2013 and one with the start year at 2017, in order to cobble together revenue projections from 2013 through 2026.

brittainhard:
@MattHJensen 

> Conforming to the goal of a unified API for reform files and GUI based inputs, I'd suggest following the same technique that is used for the asterisk (*) operator.

Not quite sure I understand this part, can you clarify a bit? Where do you see this asterisk coming into the JSON pushed forward to the server? Is it something like `{2017: "*"}`?

MattHJensen:
> Not quite sure I understand this part, can you clarify a bit? Where do you see this asterisk coming into the JSON pushed forward to the server? Is it something like {2017: "*"}?

The asterisk is not used in the JSON files. My point is just that these two are equivalent 
1. REFORM FILE: start year = 2017 &  {"policy": {"_FICA_ss_trt": {"2019": [0.1] }}:
2. TaxBrain GUI: start year = 2017 & Social Security payroll tax rate : *, *, 0.1 

So ideally their API representation is the same.

brittainhard:
This fix is pretty simple. Going to add some regression testing to ensure that first_budget_year is being sent forward (which was the source of the problem).

brittainhard:
@hdoupe did you have any luck using `cps.csv.gz` for testing?

martinholmer:
@hdoupe said in #593:
> In the future, it would be nice to have access to the raw cps file. I'm not sure how to import that from taxcalc now. Perhaps, there should be a static function in taxcalc/records.py like this:
```
    @staticmethod
    def cps_data(exact_calculations=False,
                        growfactors=Growfactors()):

        return dict(data=pd.read_csv(os.path.join(Records.CUR_PATH, 'cps.csv.gz'), compression='gzip'),
                       exact_calculations=exact_calculations,
                       gfactors=growfactors,
                       weights=Records.CPS_WEIGHTS_FILENAME,
                       adjust_ratios=Records.CPS_RATIOS_FILENAME,
                       start_year=CPSCSV_YEAR)
```
> @martinholmer I would be interested in hearing your thoughts on this.

Something like the above might be useful.  But exactly when would it be useful to have CPS data in a pandas DataFrame rather than in a Records object?

hdoupe:
@martinholmer said:
>Something like the above might be useful. But exactly when would it be useful to have CPS data in a pandas DataFrame rather than in a Records object?

After giving this some more thought, I can't think of a good reason for why we would want to use the raw data.  At first, I thought that we would need it since we need the puf data in the webapp, but I realized that we have it because it can't be in the `taxcalc` repo.  `CPS.csv.gz` can be accessed through the `Records.cps_constructor()` function.  So, unless @brittainhard can think of a reason the csv file would be needed instead of the `Records` object then we can drop this suggestion for now.

However, for testing the dropq function in the `webapp`, it may be beneficial to make the dropq functions in `taxcalc` compatible with the CPS data.  This could be as easy as adding a `use_cps` flag argument to `run_nth_year_tax_calc_model` and doing something like the following code snippet when a `Records` object is created:

```
if use_cps:
    recs1 = Records.cps_constructor(growfactors=growfactors_pre)
else:
    recs1 = Records(data=taxrec_df.copy(deep=True),
                    gfactors=growfactors_pre)
```

Or, there could be some other complications there that I don't immediately see.

martinholmer:
@hdoupe said:
> However, for testing the dropq functions [called] in the webapp, it may be beneficial to make the dropq functions in taxcalc compatible with the CPS data. 

Why?  What advantage would this provide?
Doesn't everybody testing TaxBrain have access to the `puf.csv` file?

brittainhard:
@martinholmer It would be useful for unit tests. We can set up some regression tests using the full pipeline from webapp to taxcalc. Right now I can only test the API functions in webapp.

Thats mainly why I was asking the question.

hdoupe:
@martinholmer I definitely do not have a complete understanding of how the travis testing on github works.  But, the way I understand it is that it does not use the `puf.csv` file.  Thus, if we want to test the dropq functionality using travis, then the dropq functions in `taxcalc` would have to be compatible with the `cps.csv` file.

However, I could be completely off base here.  If so, I am pair programing with @brittainhard tomorrow and maybe he can help clear up some of my confusion.

martinholmer:
@hdoupe said:
> I definitely do not have a complete understanding of how the travis testing on github works. But, the way I understand it is that it does not use the `puf.csv` file. 

That is correct.  The `puf.csv` file is not for public use, so it can never be used in GitHub tests.

> Thus, if we want to test the dropq functionality using travis, then the dropq functions in taxcalc would have to be compatible with the cps.csv file.

Why not just use the `puf.csv` file for local testing and skip tests that use the `puf.csv` file on GitHub.  That is the approach we've been taking with Tax-Calculator for years.  This is done inserting a `@pytest.mark.requires_pufcsv` decorator on the line above the `def test_...(...):` line for each test that uses the `puf.csv` file.  Then on GitHub the pytest suite is executed using:
```
py.test -m "not requires_pufcsv"
```
and all the tests that need `puf.csv` are skipped.  On your local computer, you do have `puf.csv` and so you execute:
```
py.test
```
in order to run all the tests.

Does this procedure make sense?  Can't you use this approach for TaxBrain testing?

I don't see an easy way to adapt dropq to work with the `cps.csv` file.

@brittainhard

brittainhard:
@martinholmer I like this idea. It would be a huge benefit for making the app more stable.

@PeterDSteinberg and I have had talked before about setting up a more extensive testing suite that will use the services to run actual calculations. Some of these tests can be used in `travis.yml` but it might be better to design them with local unit and integration tests in mind.

Issue #543 makes mention of a larger mocked system for testing interactions, which is good, but we could also benefit from a much larger services based test structure.

I think that should be tabled for now and we should go ahead with the tests we have now. My test covers the regression, which was failing to include the budget year value, and we should add an issue for that pytest decorator and local tests. I'll also open a larger discussion issue about ways to enhance testing beyond simple unit testing.

Issue #607, opened 2017-08-08
AMT_CG_rt is pinned to CG_rt


hdoupe:
I am creating a separate issue for the bug found in #606.

_AMT_CG_rt{1,2,3} -- If _CG_rt{i} is altered, then the corresponding rate _AMT_CG_rt{i} is pinned to the former rate, i = 1, 2, and 3. But _AMT_CG_rt4 is not affected. If _CG_rt{1,2,3} are not altered then _AMT_CG_rt{1,2,3} are correctly set to the specified values. See http://ospc-taxes7.herokuapp.com/taxbrain/edit/1406/?start_year=2017 and http://ospc-taxes7.herokuapp.com/taxbrain/edit/1407/?start_year=2017

This is probably a bug and not a feature.  @martinholmer found the relevant issues and PRs: 

> First, look at [item 4 in Tax-Calculator issue 1074](https://github.com/open-source-economics/Tax-Calculator/issues/1074) and [item 4 in Tax-Calculator pull request 1109](https://github.com/open-source-economics/Tax-Calculator/pull/1109), which was merged on December 22, 2016.
>
> Then look at TaxBrain pull request #448 (merged on January 12, 2017), which was supposed to be the fix on the TaxBrain side. However, the #448 TaxBrain changes did not eliminate the AMT_fixup logic in views.py, which had been added earlier on September 26, 2016, in pull request #342.
> 
> So, maybe I'm reading too fast, but it seems as if the behavior you've found was once considered a "feature", then became to be viewed as a "bug", but was never removed from TaxBrain. You're doing a great job on this TaxBrain debugging, so I'll leave it to you to read these issues and pull requests more carefully than I have to see what the intention was. Only then will you be able to determine if this behavior is a "feature" or a "bug".

I will open a PR with a test exposing this bug shortly.

@brittainhard @PeterDSteinberg

talumbau:
@hdoupe can you look at this issue and see if it sheds light on the problem:

https://github.com/OpenSourcePolicyCenter/webapp-public/issues/340

I'm wondering if your POST contains a parameter that is no longer in the TaxBrain GUI interface. It should still be fixed if there is a problem, but the discussion in #340 seems relevant

hdoupe:
@talumbau I'm pretty sure that it is still in the TaxBrain GUI interface.

https://www.ospc.org/taxbrain/
[...]

PR #608, opened 2017-08-08
Test AMT_CG_rti input

hdoupe:
This PR creates a test exposing the bug in issue #607.  At some point, it was desirable for `AMT_CG_rt{1, 2, 3}` to be pinned to `CG_rt{1, 2, 3}`.  In `test_views.py`, the function `test_taxbrain_rt_capital_gain_goes_to_amt` tests this feature.  In the function `test_taxbrain_amt_cg_rt_not_pinned_to_cg_rt`, this feature is treated as a bug.  Thus, the two functions are incompatible (both cannot be passed at the same time) unless `AMT_CG_rt{i}` is supposed to be pinned to `CG_rt{i}` when `AMT_CG_rt{i}` is not specified.

@martinholmer @PeterDSteinberg @brittainhard

### DIFF
--- a/webapp/apps/taxbrain/tests/test_views.py
+++ b/webapp/apps/taxbrain/tests/test_views.py
@@ -413,6 +413,69 @@ class TaxBrainViewsTests(TestCase):
         assert tsi2.AMT_CG_brk2_3 == u'451000.0'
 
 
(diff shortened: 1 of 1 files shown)


hdoupe:
Closing. No longer relevant.

Issue #609, opened 2017-08-08
_ID_Medical_frt_add4aged not accepted


hdoupe:
I am creating a separate issue for the bug found in #606.

I received a warning when I entered -0.02 for `_ID_Medical_frt_add4aged` which is in the [permitted range of -0.025 to 0.0](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/current_law_policy.json#L785).  I am able to submit after the warning.  I will open a PR with a test exposing this issue.  

@martinholmer @PeterDSteinberg @brittainhard

hdoupe:
Closed via #641

PR #610, opened 2017-08-08
Test Id medical frt add4aged

hdoupe:
This test exposes the bug in issue #609


hdoupe:
I'm closing #610. The type checking that caused this bug was removed as part of PR #641

Issue #584, opened 2017-07-18
Webapp Cleanup 


ecbrown5:
The "About" page on ospc.org needs to include Cost of Capital Calculator with logo under "Our Products."

Also on the "About" page, all three of the social media links (Twitter, Facebook,LinkedIn) are inactive.

The "Apps" link on the bottom of the home page is also inactive.

Thank you for your help!

MattHJensen:
@GoFroggyRun, could you add this to your list of issues to handle when you are back in the U.S.?

Issue #494, opened 2017-03-01
Apps Nav Links Only to TaxBrain


brittainhard:
There is no apps page to move to. Need to remove this link, or add some kind of dropdown to go to the particular app.

hdoupe:
Closed via #698 and OpenSourcePolicyCenter/blog-posts#11

PR #611, opened 2017-08-08
Web Page Clean Up

GoFroggyRun:
This PR tries to close issue #584.

1. CCC has been added to "OUR PRODUCTS":

screen shot 2017-08-08 at 12 04 41 pm

2. Twitter link has been added. (FB and LinkedIn links are still pending)

3. The "APP" link on the bottom of home page has been fixed, which directs to TaxBrain page. 

@brittainhard Please review. 

cc @MattHJensen

### DIFF
--- a/templates/pages/about.html
+++ b/templates/pages/about.html
@@ -103,7 +103,7 @@
       <section class="contact-info">
         {% flatblock "reach_out_community_link" %}
         <ul class="contact-social">
-          <li class="contact-twitter"><a href="#">Post link on Twitter</a></li>
+          <li class="contact-twitter"><a href="https://twitter.com/PolicyBrains">Post link on Twitter</a></li>
           <li class="contact-facebook"><a href="#">Share on Facebook</a></li>
           <li class="contact-linkedin"><a href="#">Discuss on LinkedIn</a></li>
         </ul>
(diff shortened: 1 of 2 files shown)


brittainhard:
@GoFroggyRun @MattHJensen I think this PR brings up the problem in #494 that I noticed a while back. The link says "apps" but only goes to TaxBrain and not CCC.

I think we might need to add some code that handles this. We can simply add two more links, one for taxbrain and one for ccc. Any thoughts?

I also noticed that the "Our Mission" part of the about page is missing.

GoFroggyRun:
@brittainhard said:

 >I think we might need to add some code that handles this. We can simply add two more links, one for taxbrain and one for ccc.

This approach makes sense to me since TaxBrain is no longer the only app we have. 

And:

>I also noticed that the "Our Mission" part of the about page is missing.

This part, along with "Our History" section, can be filled when context become available.

MattHJensen:
> I also noticed that the "Our Mission" part of the about page is missing.

When I go to http://www.ospc.org/about/, I do see content in the "our mission" and "our history" sections. 

* The content is, however, out of date. I hope to address it soon. 

> We can simply add two more links, one for taxbrain and one for ccc.

That makes sense to me.

Issue #598, opened 2017-08-02
Itemized deduction ceiling switch not correctly processed by TaxBrain GUI interface


hdoupe:
I noticed that setting the switches in `Deductions subject to the cap on itemized deduction benefits` to false did not actually cause the parameter, `_ID_BenefitCap_Switch`, that was passed to `taxcalc` to be false.  I ran this [reform](http://ospc-taxes7.herokuapp.com/taxbrain/edit/1388/?start_year=2017) on a local instance of taxbrain and the value of `_ID_BenefitCap_Switch` was passed to `taxcalc` as 

``` {... u'_ID_BenefitCap_Switch': [[True, True, False, True, 1.0, 1.0, 1.0]], ...}```

However, it should have been 
``` {... u'_ID_BenefitCap_Switch': [[1.0, 1.0, 0.0, 1.0, 0.0, 1.0, 1.0]], ...}``` or 
``` {... u'_ID_BenefitCap_Switch': [[True, True, False, True, False, True, True]], ...}```

But when I upload the reform as a file, `ID_BenefitCap_Switch` is passed back as:

```{...u'_ID_BenefitCap_Switch': [[1.0, 1.0, 0.0, 1.0, 0.0, 1.0, 1.0]], ...}```

which is what I expected.

Here is a link to the test app run of the reform file: http://ospc-taxes7.herokuapp.com/taxbrain/1389/

And here are the contents of the reform file:
```
{
    "policy": {
        "_ID_BenefitCap_Switch":
            {"2017": [[1.0, 1.0, 0.0, 1.0, 0.0, 1.0, 1.0]]},
        "_ID_c":
            {"2017": [[20000, 40000, 20000, 20000, 99e9]]}
    }
}

```
So, it appears that there is an issue with the GUI interface to Taxbrain, but the file upload functionality appears to work correctly.

@brittainhard @PeterDSteinberg

martinholmer:
@hdoupe,  Thanks for the bug report in issue #598. 
Let me be sure I understand your findings.
Are you saying that when you use the regular GUI interface to TaxBrain, the reform passed to taxcalc is correct, but when you use the file-upload interface to TaxBrain, the reform passed to TaxBrain is incorrect?  Is this right?  Or am I confused?

@brittainhard @PeterDSteinberg

hdoupe:
[...]

PR #618, opened 2017-08-11
add test causing ID_BenefitCap_switch bug

hdoupe:
This test exposes bug in #598.


hdoupe:
Closing #618. The parameter type checking that caused this bug was removed in #641.

Issue #545, opened 2017-05-03
Still more enhancements needed for TaxBrain results tables


martinholmer:
The tax analysis described in [Tax-Calculator issue 1335](https://github.com/open-source-economics/Tax-Calculator/issues/1335) reveals the need for several more enhancements to the TaxBrain results page that are not listed in issue #534.

1. When clicking on either Current Law or Reform the row labels in the Total Liabilities table don't switch to a more appropriate label. In all three versions of the table the row label ends in Change, which is appropriate in only one of the three cases. Can we drop Change from the end of each of the table row labels?

2. A key statistic is missing from the Total Liabilities tables: the ten-year totals. Most users will want the ten-year total for each row in the Total Liabilities table, but it is not there at the end of the row where most users would expect to see it. Can we add a ten-year total column on the far right?

3. Has anyone used the Excel download capability successfully? Whatever is downloaded will not import into the OpenOffice spreadsheet. I don't have Excel on my computer, does everything work smoothly when you do have Excel? I'm curious, why this option is available in an open-source project. What capability is the Excel download providing that the CSV download is not providing?

4. Print option for the Total Liabilities table has meaningless title. Can we add an informative title, like the (#534-corrected) one on the table when it is being viewed in a browser?

@MattHJensen @feenberg @PeterDSteinberg @brittainhard @enelson1995

martinholmer:
TaxBrain fixes most of the issues raised in #545.  Thanks, @GoFroggyRun for the work on this.
However, my question about the Excel download has not been addressed.
I'm going to close this issue #545 and open a new one containing that question about the Excel download.

PR #621, opened 2017-08-15
Change title name for liability table

GoFroggyRun:
Per requests in #545, the title name for liability table is not right when user exports the table via "Print" button. 

Beforehand, the title name was "Welcome to TaxBrain". After this PR, the title name, when being printed, would be more informative, that is also consistent with the title name on the result page. 

The new page generated from the "Print" button for liability table will look like:

screen shot 2017-08-15 at 11 51 44 am

@brittainhard Could you please review?

cc @MattHJensen


brittainhard:
@GoFroggyRun is this ready for review?

GoFroggyRun:
@brittainhard yep. Please take a look.

Issue #620, opened 2017-08-15
Mislabeled button on Static Results page


martinholmer:
On the TaxBrain Static Results page the first line of the Build Table section has two buttons: `Diagnostic Table` and `Difference Table`.

This issue requests that the `Diagnostic Table` be renamed as the `Distribution Table`.

This request is being made for a couple of reasons. 

First, diagnostic carries much less meaning than distribution.  Diagnostic usually refers to the process of determining what is wrong, which is not the issue here.

And second, the term diagnostic is very confusing to anybody with any knowledge of Tax-Calculator because there is a diagnostic table utility function in Tax-Calculator that has nothing to do with the information shown in TaxBrain.  In fact, what is shown in TaxBrain are the results from the Tax-Calculator `create_distribution_table` utility function.

So, why not reduce the confusion and simply change the label on the button?

@MattHJensen @GoFroggyRun

PR #626, opened 2017-08-17
Rename "Diagnostic Table" to "Distribution Table"

GoFroggyRun:
Per request in #620, one of the buttons in "build table" option is mislabeled. This PR renames "diagnostic" as "distribution" for this particular button, as well as for the tooltips in the code base. 

After this PR, the build table panel would look like the following:

screen shot 2017-08-17 at 3 43 11 pm
 
@brittainhard could you please review?

cc @martinholmer @MattHJensen

### DIFF
--- a/static/js/taxbrain-tablebuilder.js
+++ b/static/js/taxbrain-tablebuilder.js
@@ -306,7 +306,7 @@ $(function() {
                 <br>\
                 <br>\
                 <ul class="nav nav-pills nav-justified">\
-                  <li data-difference="false"><a data-toggle="tooltip" data-placement="bottom" title="<%= tooltips.diagnostic %>" href="#">Diagnostic Table</a></li>\
+                  <li data-difference="false"><a data-toggle="tooltip" data-placement="bottom" title="<%= tooltips.distribution %>" href="#">Distribution Table</a></li>\
                   <li data-difference="true" class="active"><a data-toggle="tooltip" data-placement="bottom" title="<%= tooltips.difference %>" href="#">Difference Table</a></li>\
                 </ul>\
                 <br>\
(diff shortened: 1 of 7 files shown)


martinholmer:
@GoFroggyRun,  Thanks for the changes in #626.

I'll leave it to others to review the details, but I am curious about what the difference is between `static/js/taxbrain-tablebuilder.js` and ` staticfiles/js/taxbrain-tablebuilder.js`.  Why did you need to change `Diagnostic` to `Distribution` in **both** of those files?  What's the difference between the file in the `static` directory tree and the file in the `staticfiles` directory tree?

@MattHJensen @brittainhard

martinholmer:
@GoFroggyRun,
One more question about pull request #626.  If we're changing the name on the button, shouldn't we also be changing the name of the CSV (and I guess, Excel) files that are downloaded?  Right now those filenames include the phrase `diagnostic`.  All you need to do is change `diagnostic` to `distribution` in the filenames.

GoFroggyRun:
@martinholmer asked:

>I'll leave it to others to review the details, but I am curious about what the difference is between static/js/taxbrain-tablebuilder.js and staticfiles/js/taxbrain-tablebuilder.js. Why did you need to change Diagnostic to Distribution in both of those files? What's the difference between the file in the static directory tree and the file in the staticfiles directory tree?

I might not be the best person to answer this as I am not quite familiar with the infrastructure of webbapp. All I know is that, when database mode were changed, command `python manage.py collectstatic` needs to be executed so that changes we made in `static/js/...` will be also copied to `staticfiles/js/...`. @brittainhard  should have much more output than I do.

@martinholmer also said:
>One more question about pull request #626. If we're changing the name on the button, shouldn't we also be changing the name of the CSV (and I guess, Excel) files that are downloaded? Right now those filenames include the phrase diagnostic. All you need to do is change diagnostic to distribution in the filenames.

Thanks for noticing this, I have changed the name for those downloaded file in my most recent commit [9a0d292
](https://github.com/OpenSourcePolicyCenter/webapp-public/pull/626/commits/9a0d2923b48849a15d9b2f5b3d079be1bd9c0a0b).

Issue #624, opened 2017-08-17
1.0.2 Release


brittainhard:
@hdoupe @martinholmer @MattHJensen @GoFroggyRun  I've made a milestone for the 1.0.2 release. Can we discuss what fixes and PRs we want in that milestone? I know for sure I want #612 and #589 fixed. Any other suggestions?

hdoupe:
@brittainhard what do you think about tackling #619?  I would like to fix #596, #598, and #607 as soon as possible since they will lead to incorrect results.  It appears that solving #619 will take care of #596, #598, and #607.  That way we don't have to fix the latter three bugs twice.

I'm interested in working on this if you're busy working on other stuff before the next release.

cc @martinholmer @MattHJensen @GoFroggyRun

hdoupe:
#622 should be ready to go.  This points the TaxBrain urls to the correct version of the relevant package instead of the master repo on github.  It also removes the reference to the commit ID.

hdoupe:
@brittainhard I have some tests that post reforms to TaxBrain using the requests library and compare the results with those generated by sending the reform data straight to taxcalc.  Should we try to include those in this release?

cc @martinholmer @MattHJensen @GoFroggyRun

brittainhard:
@hdoupe #619 looks like a good fix, and I think i'm the only person that can tackle it. I'll take a look.

In regards to the test PRs, those are always important to have. It would be good if you could associate that code with an issue, and add that issue to the milestone.

brittainhard:
With #622, I think its better to add issues to these milestones and not PRs. Can you open an issue for that PR?

GoFroggyRun:
@brittainhard, #611, #621 and #623 should be ready to go as well. It would be great to include them in 1.0.2 release.

brittainhard:
@GoFroggyRun great, can you add the associated issues to the milestone?

GoFroggyRun:
@brittainhard it seems that I wasn't able to associate those issues to the milestone, but related issues are #584 and #545.

brittainhard:
I added them.

brittainhard:
@GoFroggyRun did you have time to take another look at #613 ?

hdoupe:
@brittainhard I just opened an issue #625 for PR #622.

hdoupe:
@brittainhard said
> @hdoupe #619 looks like a good fix, and I think i'm the only person that can tackle it. I'll take a look.

[...]

Issue #615, opened 2017-08-11
Using requests for first iteration of automated webapp testing


hdoupe:
In this issue, I am continuing the conversation in #602 about using requests to post data to TaxBrain.  

@talumbau gave an excellent explanation for how data is posted to TaxBrain (and other websites):

> I see where you are going and this is a good idea. There is a "tool" that is missing from your proverbial toolbox that will help you achieve this goal. First I will explain what happens when someone does a TaxBrain submission:
>
> The user goes to /taxbrain and the TaxBrain screen loads. They are given a "session ID" by the site and this is kept in the browser session as long as the tab stays open. The user fills out all desired boxes and hits the submit button. This generates a specific type of message from the user's browser sent to ospc.org using the HTTP protocol. The message is called a "POST". This HTTP POST contains all of the data that the user filled out in the forms, along with the session ID. The website is designed to run specific code when it receives such a POST message. This code does all of the work of storing the user's input as a row in the taxbrain_taxsaveinputs table (a Model "instance" in Django terms), submitting the job to the worker nodes, waiting for the results, etc.
>
> All of this can, of course, be done programmatically via a script instead of typing the data in the boxes displayed in the browser. In other words, one can write Python code (or code in many other languages) that loads the /taxbrain page, but instead of displaying it in a browser, just keeps the data in a data structure. Then, one can construct an HTTP POST message with the desired parameter values (really just key-value pairs where the keys are the parameter names and values are what the user would type in the boxes) and the session ID, and then submit this POST to the site.

[...]

Issue #606, opened 2017-08-07
Version 1.0.1 Testing 


hdoupe:
I discovered several bugs in the taxbrain interface.  I used the requests library as suggested by @talumbau in issue #602 to create reforms on the test webapp and the local instance of the webapp.  I uploaded the script `post_reforms.py` on my `webapp-public` [branch](https://github.com/OpenSourcePolicyCenter/webapp-public/compare/master...hdoupe:reform_processing).  

I altered every parameter in the taxbrain interface, did a HTTP POST on my local webapp instance, copied the keywords to dropq from the celery log, and compared the expected dictionary to the actual dictionary in `hdoupe/webapp-public/reform_input_processing/gen_reforms.assert_reforms_equal()`  I think this could be better done using the functions in [`test_views.py`](https://github.com/OpenSourcePolicyCenter/webapp-public/blob/master/webapp/apps/taxbrain/tests/test_views.py) and I will spend time going over that tomorrow.

- `_ID_Medical_frt_add4aged` -- I received a warning when I entered -0.02 which is in the permitted range of -0.025 to 0.0.

- `_AMT_CG_rt{1,2,3}` -- if I alter `_CG_rt{i}` then the corresponding rate `_AMT_CG_rt{i}` is pinned to the former rate, i = 1, 2, and 3.  But `_AMT_CG_rt4` is not affected.  If `_CG_rt{1,2,3}` are not altered then `_AMT_CG_rt{1,2,3}` are correctly set to the specified values.  See http://ospc-taxes7.herokuapp.com/taxbrain/edit/1406/?start_year=2017 and http://ospc-taxes7.herokuapp.com/taxbrain/edit/1407/?start_year=2017  I may be wrong and this could be a feature and not a bug.  @martinholmer what are your thoughts here?

- As noted in issue #598 the last three switches of `_ID_BenefitSurtax_Switch` and `_ID_BenefitCap_Switch` cannot be switched to `False`.

Tomorrow, I will clean up the scripts that I posted above and set up tests for the bugs listed above.  I will also look into varying the start year and uploading reform files.  Hopefully this first round of testing will be finished by the end of the day tomorrow.

@martinholmer @PeterDSteinberg @brittainhard

hdoupe:
[...]

Issue #602, opened 2017-08-03
Testing Improvements


brittainhard:
I think it would be good to have a discussion about testing strategies that would go beyond unit tests and mocking functions. One goal for issue #543 was to add a mocked test, much of which we have in the `MockCompute` object.

I think we should also talk about creating unit tests that run the actual services and test the output. These would only be run locally, since they would require the taxpuf file.

Issue #600  is one we can get done quickly, and the issue in #601 needs to be dealt with.

@martinholmer @MattHJensen @hdoupe @PeterDSteinberg if you have any suggestions I think this is a good place to discuss them.

hdoupe:
I would like to create a test that feeds a reform (or a series of reforms) through the interface that alters all of the parameters and validates that the inputs are processed as expected.  However, the process of inputting over a hundred parameters into taxbrain takes several hours and is painfully tedious.

I've been trying to think of a way to come as close as possible to doing this without spending several hours punching values into the interface.  Here are some of my ideas:

1. If you input values directly into the interface, then this page is saved as something like this: https://www.ospc.org/taxbrain/edit/1000/?start_year=2015.  But, if you feed a json file into taxbrain then an input page is not created and saved.  So, what if we could automate the loading of a json file into taxbrain, save the parameters, and then have a page similar to the one occurring if we input the parameters manually?

2. The html source code is not available in the production app, but it is available in the test app.  What if we used something like selenium to programmatically input the parameters into the taxbrain interface page?

3.  The pages like the link I listed in (1) are saved in the `taxbrain_taxsaveinputs` table.  What if we wrote a script that takes a reform as an input, reads the parameters, and directly adds a row to `taxbrain_taxsaveinputs`?  Then we would have a link similar to the one referenced in (1).  We could then go to that link and run the parameters through the taxbrain model.

The idea is to figure out a way to simulate as closely as possible the act of feeding user inputs through taxbrain and checking the results without actually doing it.  Of course, I'm open to other ideas and the three that I listed may not be very good.  This is something that we could start thinking about.

The next step would be to create a keywords to dropq dictionary like the one created in the `celery_tasks.dropq_task` function.  This dictionary is the expected output from feeding the reform through the interface.  If the actual output and expected output are the exact same (down to the type of the objects), then the dropq results would be identical.  This we could test in the same way as my code snippet [here](https://github.com/open-source-economics/Tax-Calculator/issues/1485#issuecomment-318689633).

[...]

PR #627, opened 2017-08-18
[WIP] Test reform post and results

hdoupe:
This PR mainly addresses issue #615 but was brought up in #602 and #606. It adds a series of scripts that posts a reform file or a reform through the Taxbrain GUI on a local Taxbrain instance using the requests library and simultaneously runs the same reform directly using `taxcalc.dropq.run_nth_year_tax_calc_model`.  The results from TaxBrain are retrieved from the local sqlite3 table `taxbrain_taxsaveinputs` and the results from submitting directly to `taxcalc` are formatted into the same structure as the TaxBrain results.  These two results are then compared.  

This effectively tests two parts of TaxBrain:

1. The reform data is processed as expected by TaxBrain.  If it is not, then the end results will not be the same.
2. The process of sending the processed parameters and receiving the correct results from `taxcalc` works as expected.

Currently, it takes about 37 minutes to run the 7 tests (each runs for about 5 minutes).  I think that the number of tests could be decreased.  For example, I don't think the tests involving the Trump and RyanBrady plans test anything that is not tested using r1.json and the reform which posts all of the taxbrain parameters.  

The TaxBrain GUI parameters are changed in unexpected ways such as being rounded, casted from floats to integers (1.0  to 1), and from boolean to binary (0-1).  Since the dropq algorithm generates a random seed directly from the reform data, this unexpected data manipulation results in using different data sets to estimate the results from using the same reform.  Thus, to test the TaxBrain GUI, I tried to compare the reform results within a 2% tolerance.

Further, it does not work when the entire test suite is run.  I think that at some point in the test suite the puf file and maybe the requests functions are mocked.  To run these tests run `py.test webapp/apps/taxbrain/tests/reform_input_processing/` and make sure the server and supervisord are running in other terminals.  To run the test suite without these tests run `py.test webapp/apps/ -m "not requires_pufcsv`

Initially, I hoped that this PR would be ready by the 1.0.2 release (#624).  However, I don't think it will be ready in time.  I think it could still be cleaned up a good bit and maybe some of the functions could be mocked.  

cc @martinholmer @brittainhard


hdoupe:
Closing PR #627. See rationale in #615

PR #637, opened 2017-08-23
Handle data from new BTAX Version

### LINKED ISSUE #634: Using BTAX DataFrames for Dynamic Plotting
@hdoupe @MattHJensen @PeterDSteinberg

The central problem is that the data returned from BTAX to CCC cannot be used for plotting -- specifically the plots that Haylee and Dora have created.

BTAX serializes the original DataFrames for use in the HTML table, but it doesn't preserve the original structure of the many DataFrames that BTAX returns. They need to change the BTAX code to return serialized versions of those DataFrames alongside the JSON to be used for the HTML table.

If they can have access to the original DataFrames, then they can use them to create dynamic plots to be put onto the CCC results page alongside the HTML table.

They need to change the BTAX function to return the DataFrames, and we need to include that new version into our next release. Their deadline is the end of this week.

brittainhard:
Closes #634


hdoupe:
Thanks for the contribution @brittainhard. Can you update B-Tax to the new version, 0.1.9?

brittainhard:
@hdoupe you mean in the requirements file?

brittainhard:
@hdoupe I merged your code into this branch and found a bug. Looks like btax is relying on the old dropq code to submit, since its still using form data to submit. I can have this fixed quickly I think.

hdoupe:
@brittainhard thanks for catching this.  The only validation that the webapp does on TaxBrain data is to make sure that no extra parameters were submitted and all of the data parses.  I think we could leave all of the form code as it was before #641 but add a taxbrain flag that tells the validation routine to skip over the min/max validation in [`taxbrain/forms.do_taxcalc_validations`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/2e781cf9e4091ce8716286f8e11c6aaa80185b85/webapp/apps/taxbrain/forms.py#L251).

brittainhard:
@hdoupe i think the problem is with how webapp is now interacting with btax. I'm getting this error from the backend: `Error: runner_json_tables() argument after ** must be a mapping, not float`, in this file here: https://github.com/open-source-economics/B-Tax/blob/master/btax/run_btax.py.

I think we need to allow for btax to submit reforms in the old way.

hdoupe:
@brittainhard I opened a PR ( #651 ) on the 634_btax_dataframes branch.  Is that you what you were asking me to do?

brittainhard:
@hdoupe yep, thats what i was looking for, thanks.

PR #646, opened 2017-09-01
Add summary row for table

brittainhard:
@GoFroggyRun this is a dirty prototype but it works. It just loops through the tables of the existing data structure and adds up their values, then creates a new column. The front-end is handling it well, no changes need to be made there. I haven't tested this solution thoroughly, and the code could be improved. This is a good basis for the feature I think. My output from a reform looks like this:

screen shot 2017-09-01 at 1 44 29 pm

It's adding the column AFTER the table has already been saved in the database. I think this can also be done after before the data is saved in the database. I don't think that will cause issues with backwards compatibility since the javascript table is just accepting arbitrary data as long as its in the right format.

@martinholmer @hdoupe your input on this would he helpful I think.

### DIFF
--- a/webapp/apps/taxbrain/views.py
+++ b/webapp/apps/taxbrain/views.py
@@ -481,6 +481,25 @@ def edit_personal_results(request, pk):
     return render(request, 'taxbrain/input_form.html', init_context)
 
 
+def add_summary_column(table):
+    import copy
+    summary = copy.deepcopy(table["cols"][-1])
+    summary["label"] = "Summary"
+    table["cols"].append(summary)
+    table["col_labels"].append("Summary")
+    for x in table["rows"]:
+        row_total = 0
+        for y in x["cells"]:
+            row_total += float(y["value"])
+        x["cells"].append({
+            'format': {u'decimals': 1, u'divisor': 1000000000},
+            u'value': unicode(row_total),
+            u'year_values': {}
+        })
+    return table
+
+
+
 def get_result_context(model, request, url):
     output = model.tax_result
     first_year = model.first_year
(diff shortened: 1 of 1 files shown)


martinholmer:
@brittainhard,  I think it would be far better to label the new column `Total` (instead of `Summary`).

hdoupe:
@brittainhard said
> It's adding the column AFTER the table has already been saved in the database. I think this can also be done after before the data is saved in the database. I don't think that will cause issues with backwards compatibility since the javascript table is just accepting arbitrary data as long as its in the right format.

I think that this should probably be done before the model is saved to the database.  That would make more sense to me.  However, I am not very familiar with java script.  So I have no comments on the backwards compatibility of this.

martinholmer:
@brittainhard, Thanks for the quick response.

brittainhard:
@hdoupe that makes sense, but if we were to need to make changes to the column, or even add another column, the older reforms wouldn't have this up-to-date structure.

GoFroggyRun:
@brittainhard It seems to me that having this number after the database shouldn't be a problem. Javascript should be able to take care it. 

Thanks for taking care of this enhancement. I will work on some refactorizations after this PR being merged.

brittainhard:
@GoFroggyRun I don't think you're going to have to change any of the javascript. That's the nice part of this approach. I want to change the style of the code I wrote and probably refactor it, but then I think it will solve the issue.

Is this a high priority issue?

martinholmer:
@brittainhard asked in #646:
> Is this a high priority issue?

**Yes!**  This enhancement request dates back May 3, 2017 (see issue #545).

PR #653, opened 2017-09-18
Remove submit_micro function

hdoupe:
This PR removes the `taxbrain/views.submit_micro` function.  It appears that this function would be called to submit a reform after the user submitted the reform using the quick-calc option.  However, the `edit_personal_results function` is now used for this.


hdoupe:
I now know exactly what `submit_micro` is used for.  See #654

PR #655, opened 2017-09-20
Refactor submit_reform for quick-calc submission case

### LINKED ISSUE #654: Submit simulation after quick-calc bug
In PR #641, the new GUI processing logic was not implemented for the case where someone submits a quick-calculation and then submits a full-calculation from the quick-calc results page.

hdoupe:
Closes #654

### DIFF
--- a/webapp/apps/taxbrain/views.py
+++ b/webapp/apps/taxbrain/views.py
@@ -297,16 +297,83 @@ def get_reform_from_gui(request, taxbrain_model=None, behavior_model=None,
     return (reform_dict, assumptions_dict, "", "", errors_warnings)
 
 
+def save_model(url, request, model, has_errors, start_year,
+               do_full_calc, is_file, reform_dict, assumptions_dict,
+               reform_text, assumptions_text, submitted_ids,
+               max_q_length, user):
+    """
+    Save user input data
+    returns OutputUrl object
+    """
+    job_ids = denormalize(submitted_ids)
+    json_reform = JSONReformTaxCalculator()
+    # save file_input user params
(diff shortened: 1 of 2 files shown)


hdoupe:
@brittainhard PR #655 is ready for review.

brittainhard:
This is working. @hdoupe why the large refactoring job. What does the `save_model` function do that `submit_reform` wasn't?

hdoupe:
@brittainhard `submit_micro` was using the `process_model` function which was replaced with `submit_reform`. But `submit_reform` starts with a `request` object and not a `TaxSaveInputs` object.  It seemed easier to refactor the `submit_reform` function than use a `request` object as an input to `submit_reform` some of the time and a `TaxSaveInputs` object the other times.  Does that make sense?

webapp/apps/taxbrain/views.py
```diff
@@ -602,9 +632,49 @@ def submit_micro(request, pk):
     model.tax_result = None
```
hdoupe: The data is already parsed into a `TaxSaveInputs` object so that means that there is no need to go through the `request` parsing logic in `submit_reform`.

brittainhard:
@hdoupe need to fix these merge conflicts.

hdoupe:
@brittainhard Sorry about that. I should have checked this earlier.  I'll run this locally and make sure it works.

hdoupe:
@brittainhard Just ran it locally.  I submitted a quick-calc and a full calculation from there and everything looks good.

brittainhard:
LGTM, merging.

PR #665, opened 2017-09-22
Revert switches from 1/0 to True/False in model

### LINKED ISSUE #659: Itemized Deduction switches all off on edit parameters page
After the user submits a calculation and goes back to edit parameters, all itemized deduction switches are set to off.  However, there are not any issues with passing the switch parameters to `taxcalc` for the initial calculation.

hdoupe:
Closes #659


brittainhard:
This changes really needed.

PR #641, opened 2017-08-29
Gui input processing

hdoupe:
This PR addreses issue #619.  Thanks to @martinholmer's excellent work on the Tax-Calculator reform processing functions ([Tax-Calcuator/PR#1502](https://github.com/open-source-economics/Tax-Calculator/pull/1502), [Tax-Calculator/PR#1503](https://github.com/open-source-economics/Tax-Calculator/pull/1503), [Tax-Calcuator/PR#1505](https://github.com/open-source-economics/Tax-Calculator/pull/1505)), the TaxBrain GUI variables can easily be mapped to the Tax-Calculator variables.  This simplifies the processing of these parameters in TaxBrain considerably.  

In this PR, I operated under the assumption that almost all of the processing of the parameters would be done in the Tax-Calculator.  So, the only validation that is done is to make sure that no extra parameters are posted and that only valid data types are accepted (either floating point, integers, or boolean).  

The first commit in this PR moves to the new parameter processing methods but makes the minimum amount of changes required to do this.  In the second commit, some of the no longer needed code is removed.  The third commit adds some comments.

[EDIT: I had to do some back tracking.  The first commit edits the `personal_results` function so that all of the parameter processing is done there.  The second commit moves the function converting the fields parameters to json reform style parameters to `helpers.py`.  This function is named `to_json_reform` and is called from the `process_model` function.  The third commit pins the webapp to Tax-Calculator version 0.10.0.]

TODO List:
- [x] Move `minimal_processing` function to forms.py and save parameter values for user-edit functionality
- [x] Add processing of checkbox parameters (ID_c, ID_c_medical, ALD_InvInc_ec_base_RyanBrady, etc)
- [ ] Audit some of the back-end code to see what is still needed
- [x] Update tests
- [x] Handle errors

@brittainhard @MattHJensen

PR #670, opened 2017-09-27
Fix dynamic simulation button logic for file upload

### LINKED ISSUE #669: Dynamic simulation button not displayed after file input submission
The dynamic simulation button is not shown when a reform is submitted through the file-upload interface.  The dynamic simulation button should not be shown when an assumptions file is uploaded, but it should be shown when no assumptions file is uploaded.

hdoupe:
Closes #669

### DIFF
--- a/deploy/taxbrain_server/celery_tasks.py
+++ b/deploy/taxbrain_server/celery_tasks.py
@@ -70,7 +70,6 @@ def dropq_task(year, user_mods, first_budget_year, beh_params, tax_data):
                     user_mods[reform_year].pop(key)
         user_reform = {"policy": user_mods}
     print('user_reform', user_reform, user_mods)
-    reform_style = [True if x else False for x in user_reform]
     if beh_params:
         for x, y in beh_params.items():
             for z in y.keys():
@@ -93,8 +92,7 @@ def dropq_task(year, user_mods, first_budget_year, beh_params, tax_data):
                'mX_bin': mX_bin_i, 'df_bin': df_bin_i, 'pdf_bin': pdf_bin_i,
                'cdf_bin': cdf_bin_i, 'fiscal_tot_diffs': fiscal_tot_i,
                'fiscal_tot_base': fiscal_tot_i_bl,
-               'fiscal_tot_ref': fiscal_tot_i_ref,
-               'reform_style': reform_style}
+               'fiscal_tot_ref': fiscal_tot_i_ref}
 
 
     #Add taxcalc version to results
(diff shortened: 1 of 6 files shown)


brittainhard:
LGTM, I wonder how this got broken in the first place?

deploy/taxbrain_server/celery_tasks.py
```diff
@@ -70,7 +70,6 @@ def dropq_task(year, user_mods, first_budget_year, beh_params, tax_data):
                     user_mods[reform_year].pop(key)
         user_reform = {"policy": user_mods}
     print('user_reform', user_reform, user_mods)
-    reform_style = [True if x else False for x in user_reform]
     if beh_params:
```
hdoupe: `reform_style` can only be `[True]` since it is a dictionary with just one key `policy`.  Further, we now pass back assumptions data with empty dictionaries even if no assumptions are set.  So, this method no longer detected whether the user specified assumptions or not.

webapp/apps/taxbrain/views.py
```diff
@@ -766,9 +761,9 @@ def get_result_context(model, request, url):
 
     if (model.json_text is not None and (model.json_text.raw_reform_text or
        model.json_text.raw_assumption_text)):
-        reform_file_contents = model.json_text.reform_text
+        reform_file_contents = model.json_text.raw_reform_text
         reform_file_contents = reform_file_contents.replace(" ","&nbsp;")
-        assump_file_contents = model.json_text.assumption_text
+        assump_file_contents = model.json_text.raw_assumption_text
         assump_file_contents = assump_file_contents.replace(" ","&nbsp;")
```
hdoupe: After #641, all reforms have attribute `assumption_text` as an assumption dictionary but the values are empty if they are not specified via an assumptions file or the dynamic behavior interface.  So, `raw_assumption_text` was used instead to distinguish file-upload reforms with an assumptions file from other reforms.

hdoupe:
@brittainhard I had trouble pinning down the exact cause.  But I found another bug once I dug further into the code.

The `results.html` file checked two different conditions before it handled the dynamic case.  It set up the link to dynamic simulations: `allow_dyn_links` and `assump_file_contents` which is `True` if `assump_file_contents` is a string of length greater than one.  Work in #641 and later PR's changed these attributes and they no longer indicated whether the user specified the assumptions or not.  

It turned out that `allow_dyn_links` was always true and a link was created to a dynamic simulation even in cases where an assumptions file was uploaded and the "Dynamic Simulation" button was not displayed.  Also, using the absence of assumption file contents in the java script no longer worked (still not sure why).  So, I checked that condition in `views.py` and stored the value in `allow_dyn_links`.  That works ok.

Issue #596, opened 2017-07-28
EITC phase-in rate values being changed


hdoupe:
This issue was initially raised in [Tax-Calculator issue #1485](https://github.com/open-source-economics/Tax-Calculator/issues/1485#issuecomment-318723224).  

@MattHJensen said
>A TaxBrain user is wanting to set an EITC rate (eitc_rt) over 100 percent.
>
> The EITC rate description states that the range of the parameter is between 0 and 100 percent.
> 
> To experiment, I ran two reforms, one with the rate set to 100 percent for all filing statuses and one with the rates at 150 percent, and I got identical results for both.
> 
> All EITC rates at 100 percent.
> http://www.ospc.org/taxbrain/15189/
> 
> All EITC rates at 150 percent.
> http://www.ospc.org/taxbrain/15190/
> 
> - can we allow rates over 100 percent?
> 
> - was this a silent error, and are there other scenarios where that might be happening?

Here are results from the celery log when I ran the two reforms mentioned above. Celery log for `_EITC_rt` set to 1:
```
[2017-07-28 10:16:35,533: WARNING/MainProcess] keywords to dropq
[2017-07-28 10:16:35,534: WARNING/MainProcess] {'start_year': 2017, 'user_mods': {u'growdiff_response': {}, u'consumption': {}, u'growdiff_baseline': {}, u'behavior': {}, u'policy': {2017: {u'_EITC_rt': [[1.0, 1.0, 1.0, 1.0]], u'_NIIT_PT_taxed': [False], u'_ID_BenefitCap_Switch': [[True, True, True, True, 1.0, 1.0, 1.0]], u'_ALD_InvInc_ec_base_RyanBrady': [False], u'_EITC_indiv': [False], u'_ID_BenefitSurtax_Switch': [[True, True, True, True, True, True, True]], u'_II_no_em_nu18': [False], u'_CG_nodiff': [False], u'_CTC_new_refund_limited': [False]}}, u'gdp_elasticity': {}}, 'year_n': 0}
```
and `_EITC_rt` set to 1.5:
```
[2017-07-28 10:32:22,549: WARNING/MainProcess] keywords to dropq
[...]

PR #674, opened 2017-10-02
Add test for eitc rate greater than 1

hdoupe:
This test exposes the bug found in issue #596

### DIFF
--- a/webapp/apps/taxbrain/tests/test_views.py
+++ b/webapp/apps/taxbrain/tests/test_views.py
@@ -488,6 +488,36 @@ class TaxBrainViewsTests(TestCase):
         self.failUnless(response.url[:link_idx+1].endswith("taxbrain/"))
 
 
+    def test_taxbrain_eitc_rt_greater_than_1(self):
+        """
+        Transfer over the regular tax capital gains to AMT
+        """
+        #Monkey patch to mock out running of compute jobs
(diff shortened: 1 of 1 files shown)


martinholmer:
@hdoupe,  Why are you adding a TaxBrain test to detect a perfectly valid parameter value?
Look at `taxcalc/current_law_policy.json` to see that `eitc_rt` can have **any** non-negative value.
What am I missing?  I don't see why this TaxBrain test is needed.  It seems wrong, not just unneeded.

hdoupe:
@martinholmer I added this as a regression test.  I wanted to make sure that we didn't have any issues in the future with EITC_rt parameter values or values for similar parameters being rounded down to 1.

martinholmer:
Oh, the test is to see if values of `_EITC_rt` greater than one are **allowed**.
OK, I missed that.  So, adding the test seems fine.

But I don't understand the new test's docstring:
```
    def test_taxbrain_eitc_rt_greater_than_1(self):
        """
        Transfer over the regular tax capital gains to AMT
        """
```

hdoupe:
@martinholmer Thanks for pointing that out.  I copied and pasted from another function and forgot to change the docstring.  The latest commit updates the docstring.

hdoupe:
#674 test is covered in PR #706

PR #676, opened 2017-10-02
Test full calculation submission from quick-calc results page

hdoupe:
This PR adds tests to cover the gap in test coverage that was exposed by issue #654.


hdoupe:
@talumbau Thanks for reviewing.  I agree with you.  I had an issue with the `NUM_BUDGET_YEARS` environment variable getting out of sync between the value set in `test_views.py` and `compute.py`.  In some files, such as [`taxbrain/tests/utils.py`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/webapp/apps/taxbrain/tests/utils.py#L6), the environment variable is set to '2'.  It looks like it was set to '2' after some of the modules were imported.  I think this is what caused the issue.

The latest commit gets `NUM_BUDGET_YEARS` before the `compute.py` classes are imported.  That seems to have solved the problem.

brittainhard:
LGTM, merging.

hdoupe:
Thanks @brittainhard

PR #642, opened 2017-08-30
Bubble plot visualization

hayleefay:
This PR includes a bokeh bubble plot that will appear on the results page of the ccc.

PR #687, opened 2017-10-05
Bokeh Requirement in Correct File

brittainhard:


hdoupe:
After updating Bokeh to the most recent version 0.12.9, the Cost-of-Capital calculator loads but displays an [error message](https://www.ospc.org/ccc/881/).  I tested this on my local machine and CCC worked fine using Bokeh 0.12.7, but after upgrading to 0.12.9, I received an error message:
btax_localrun_1

One option is to try pinning Bokeh to 0.12.7.  Admittedly, this is a sub-optimal solution.  However, it would be a top priority to change the plot to move Bokeh up to the most up-to-date version before the next release.

@MattHJensen @martinholmer @brittainhard @hayleefay

martinholmer:
@hdoupe,  See the discussion about bokeh instability in pull request #642.
The bokeh package seems pretty flaky in the last few months with major regressions when new versions are introduced.  Why did we adopt bokeh when there is a much more stable and full-featured graphics options?

@MattHJensen @GoFroggyRun  @brittainhard

hayleefay:
@hdoube Most likely updating the version in [these link and script tags for the Bokeh widgets](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/fcee534d45890183b57284a52c747e2589903c17/templates/btax/results.html#L12-L15) will solve the problem. But, in the long run this will keep happening if we do not pin a specific version, since those CDNs are pinned to a specific version. I tried to find a way around that but the updating CDN tags don't seem to be available for the widgets.

hayleefay:
@hdoupe And of course as soon as I say that I figure out what to do. Should I put up a pull request? Or I can just tell you where to add two lines of code if you like.

hdoupe:
@hayleefay Great, please open a pull request.

brittainhard:
@hayleefay @hdoupe I think the best thing to do is pin the bokeh version. Bokeh is still in activate development and changes frequently.

hayleefay:
@brittainhard @hdoupe The PR I just submitted allows the visualization to work on 0.12.9 and doesn't expressly need a pinned version anymore, but I agree with the decision to pin a version due to development concerns.

hdoupe:
@brittainhard @hayleefay I agree that we should pin it, too.  Throwing in a quick fix like this sort of short-cuts the release process.  I'm new to managing this process, and I want to make sure that we establish good habits early.  So, let's pin the release to 0.12.7 and slate PR #688 for the next release.

@hayleefay thanks for the quick response on this issue.

martinholmer:
@hdoupe said:
> I agree that we should pin it, too. Throwing in a quick fix like this sort of short-cuts the release process. I'm new to managing this process, and I want to make sure that we establish good habits early. So, let's pin the bokeh to 0.12.7 

This is an OK approach only if it is a short-term fix.  The versions should ideally be `package>=x.y.z` not
`package==x.y.z`.

I thought the pull request submitted by @hayleefay made the charts work with with both 0.12.7 and 0.12.9.
Is that not true?  What about bokeh 0.12.8?  Do the charts work with that version?

hayleefay:
@martinholmer, my latest PR does work with both 0.12.7 and 0.12.9. Barring breaking changes in future versions, a version of bokeh>=0.12.7 will be just fine. Oddly enough there does not appear to be a 0.12.8.

martinholmer:
@hayleefay said in #687:
> my latest PR does work with both 0.12.7 and 0.12.9. Barring breaking changes in future versions, a version of bokeh>=0.12.7 will be just fine.  Oddly enough there does not appear to be a 0.12.8.

OK, thanks for the helpful information.  So, why isn't PolicyBrain using `bokeh>=0.12.7`?
I don't understand what the need for the `==` pinning.

@brittainhard and @hdoupe, can you please explain?  Are there thinking other than the charts by @hayleefay that fail when using bokeh 0.12.9?

@MattHJensen

MattHJensen:
@martinholmer said:

> This is an OK approach only if it is a short-term fix. The versions should ideally be package>=x.y.z not package==x.y.z.

@martinholmer, could you explain why you believe >= is preferable to ==? And what are the conditions that need to be met for a project to safely handle >= pinning?

Issue #686, opened 2017-10-05
CCC hanging while calculations are running


MattHJensen:
CCC is not generating results for me. I decreased the top business tax rate to .3. 

http://www.ospc.org/ccc/880/

cc @hdoupe

hdoupe:
Thanks for opening an issue.  @brittainhard and I are trying to fix it now.

hdoupe:
The test app was running Bokeh version 0.12.7 but the production app was pinned to Bokeh 0.12.5.  This was an issue that we had in the test app deployment.  This should be resolve by PR #687

MattHJensen:
Do you think it would make sense to add a test to #687 to try to catch this type of test in the future or alternatively to open a new issue for adding such a test?

martinholmer:
@hdoupe said:
> The test app was running Bokeh version 0.12.7 but the production app was pinned to Bokeh 0.12.5. This was an issue that we had in the test app deployment. This should be resolve by PR #687.

But if you just moved it (rather than put it in both files) won't the test app not work?
I guess another way of expression my confused ignorance is by asking this question:
What is the role of the `deploy/fab/dropq_environment.yml` file (which no long has a bokeh requirement)?

@MattHJensen @brittainhard

hdoupe:
@brittainhard and I discussed making the environment setup files clearer.  Upon doing this, there should definitely be tests to make sure we are running the correct versions of packages.

@martinholmer I'm not sure why they are in different files.

PR #689, opened 2017-10-06
Pin Bokeh to 0.12.7

hdoupe:
This PR pins Bokeh to version 0.12.7.  See PR #687 and issue #686 for further discussion.


hdoupe:
Looks like PR #689 resolves #686 . See https://www.ospc.org/ccc/882/

PR #729, opened 2017-11-03
Fix "news.ospc.org" link bug

### LINKED ISSUE #727: Production app hangs on clicking news.ospc.org
The production app does not load when you click the news tab.  When you do this, you select https://news.ospc.org but in the test app when you click the news tab it directs you to http://news.ospc.org.  In the latter case, news.ospc.org loads just fine.

GoFroggyRun:
This PR closes #727. 

The Policy Brain app is not able to find `news.ospc.org` without supplying a full hyper link, for the reason that the blog (news) page is hosted somewhere else outside the app. 

There's no way for me to test on my local (the previous link works just fine on local as well as on test app). I used console to substitute HTML elements on [ospc.org](https://www.ospc.org/) by hard-coding, and it seems that, after making the following changes, the website has no problem accessing `news.ospc.org`.

@hdoupe @MattHJensen

### DIFF
--- a/templates/pages/home.html
+++ b/templates/pages/home.html
@@ -68,7 +68,7 @@
                 </ul>
             </li>
             <li class="{% if section.active_nav == 'news' %}current{% endif %}">
-              <a href="//news.ospc.org">News</a>
+              <a href="http://news.ospc.org/">News</a>
             </li>
             <li class="{% if section.active_nav == 'about' %}current{% endif %}">
               <a href="/about">About OSPC</a>
--- a/templates/pages/home_content.html
+++ b/templates/pages/home_content.html
@@ -132,13 +132,13 @@
   <section class="news-items news-items-dark">
     <div class="container">
       <p class="text-right">
-        <a href="//news.ospc.org" class="see-all">See all news</a>
+        <a href="http://news.ospc.org/" class="see-all">See all news</a>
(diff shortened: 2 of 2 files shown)


hdoupe:
@GoFroggyRun This seems like it would fix the problem. But, it's tough to tell if this solves the problem since it worked on the test app before.  I can give it a try and see if it works. What do you think?

GoFroggyRun:
@hdoupe said: 
>But, it's tough to tell if this solves the problem since it worked on the test app before.

Indeed, that's why I tried to test the changes in this PR using [ospc.org](https://www.ospc.org/). 
It's still hard, however, to guarantee that this approach will resolve the problem. Let's give it a try first.

hdoupe:
@GoFroggyRun I agree that we should give it a try.  I'll do a deploy tomorrow and see what happens.

GoFroggyRun:
@hdoupe sounds good. Thanks!

hdoupe:
No problem.  Thanks for providing the quick fix.  

I think a better long-term fix would be to get https://news.ospc.org to work.  However, there are some extra steps you have to go through to do so with S3.  See this article for example: https://medium.com/@sbuckpesch/setup-aws-s3-static-website-hosting-using-ssl-acm-34d41d32e394

PR #745, opened 2017-11-20
Collapse in all fields temporarily 

GoFroggyRun:
This PR collapse in all fields temporarily @hdoupe.

### DIFF
--- a/templates/taxbrain/default_input.html
+++ b/templates/taxbrain/default_input.html
@@ -14,10 +14,10 @@
         data-target="#{{ title|make_id }}-collapse"
         aria-expanded="false"
         aria-controls="{{ title|make_id }}-collapse">
-        <span class="glyphicon glyphicon-plus" style="font-size:2em; margin-bottom:5px"></span>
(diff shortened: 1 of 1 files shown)


hdoupe:
@GoFroggyRun it looks like all of the fields are expanded instead of collapsed.  Is that what you intended?

MattHJensen:
For those working on this, the desired behavior is:

1. On ospc.org/taxbrain everything should be collapsed. 
2. On a preset input page (ospc.org/taxbrain/edit/*/, the sections with edited parameters should be open by default and others should be closed. 

The preferred suboptimal outcome is for everything to always be open rather than to achieve (1) but not (2).

GoFroggyRun:
As @MattHJensen pointed out, a temporary solution would be to expand everything for the moment, before we have figured out the cause of the "collapse-expand" issue.

Speaking of causes, after looking at the front-end code, I feel like there's nothing wrong, nor changed, in the logic that takes care of the collapse-expand functionality. One possible reason I could think of that might have caused the issue is that TaxBrain interface mistakenly regards those newly-added parameters being somehow edited, and thus expanding them. @hdoupe would that be possible?

PR #751, opened 2017-11-21
Change job_proc_time from 30 to 50 seconds

### LINKED ISSUE #750: TaxBrain reform hanging
image

When I run this simulation, it hangs on "a few seconds left" 

@hdoupe @GoFroggyRun

hdoupe:
Resolves #750 

50 seconds was chosen since the webapp takes about 40 seconds to run one year of static analysis.  This should be a safe amount of time to allow for one year's calculation.


hdoupe:
This is running successfully on the test app.

PR #753, opened 2017-11-21
Update RELEASES.md for 1.2.0

hdoupe:
If you contributed to PB 1.2.0, take a look at the README.md and make sure you agree with how your contribution is described.  Feel free to propose a different description or propose that a contribution should be moved to another category (i.e. PR 123 really should be a minor change and not a bugfix).

cc @GoFroggyRun @hayleefay @MattHJensen


hayleefay:
@hdoupe look good to me!

PR #756, opened 2017-11-30
Add Code of Conduct

hdoupe:
In this PR, I propose adding a Code of Conduct for PolicyBrain.  The goal is to ensure that this environment remains open and welcoming for contributors from all walks of life and levels of technical expertise.  From following threads in other projects and reading about successful open source projects, I think that having a Code of Conduct where misbehaving or simply frustrated members can be referred is an easy, painless way to keep discussions from devolving into ad hominem attacks or other non-productive means of debate.

The Code of Conduct proposed here was copied from [pandas](https://github.com/pandas-dev/pandas-governance/blob/master/code-of-conduct.md).  I welcome and encourage all feedback on this document.  

If we decide to keep the working group, then I volunteer myself to be in it.  If you are interested in being in this working group, then please indicate this by leaving a comment below.  Ideally, we would have 2-4 members in this group.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,59 @@
+# Contributor Code of Conduct
+
+As contributors and maintainers of this project, and in the interest of
+fostering an open and welcoming community, we pledge to respect all people who
+contribute through reporting issues, posting feature requests, updating
+documentation, submitting pull requests or patches, and other activities.
+
+We are committed to making participation in this project a harassment-free
+experience for everyone, regardless of level of experience, gender, gender
+identity and expression, sexual orientation, disability, personal appearance,
+body size, race, ethnicity, age, religion, or nationality.
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery
+* Personal attacks
+* Trolling or insulting/derogatory comments
+* Public or private harassment
+* Publishing other's private information, such as physical or electronic
+  addresses, without explicit permission
+* Other unethical or unprofessional conduct
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+By adopting this Code of Conduct, project maintainers commit themselves to
+fairly and consistently applying these principles to every aspect of managing
+this project. Project maintainers who do not follow or enforce the Code of
+Conduct may be permanently removed from the project team.
+
+This Code of Conduct applies both within project spaces and in public spaces
+when an individual is representing the project or its community.
+
+A working group of community members is committed to promptly addressing any
+reported issues. The working group is made up of pandas contributors and users.
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting the working group by e-mail (email@email.com).
+Messages sent to this e-mail address will not be publicly visible but only to
+the working group members. The working group currently includes
+
+- ...
+
+All complaints will be reviewed and investigated and will result in a response
+that is deemed necessary and appropriate to the circumstances. Maintainers are
+obligated to maintain confidentiality with regard to the reporter of an
+incident.
+
+This Code of Conduct was created by [pandas][pandascoc] developers, who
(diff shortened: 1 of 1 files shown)


Amy-Xu:
Looks good! It would be great if we have something similar for the Tax-Calculator.

martinholmer:
@Amy-Xu said:
> Looks good!

Yes, but this is not yet complete, right?

Do we really want this in a PolicyBrain code of conduct?
```
 +A working group of community members is committed to promptly addressing any
 +reported issues. The working group is made up of pandas contributors and users.
 +Instances of abusive, harassing, or otherwise unacceptable behavior may be
 +reported by contacting the working group by e-mail (email@email.com).
 +Messages sent to this e-mail address will not be publicly visible but only to
 +the working group members. The working group currently includes
 +
 +- ...
```

hdoupe:
@martinholmer thank you for contributing to this discussion.  The Code of Conduct is not complete, and this is the type of discussion that I was hoping for.

I think that a working group would be helpful even though I doubt that it would be used often.  My understanding of its role is that this group is there to guarantee that a person's complaints about the conduct of other contributors or cultural problems within the project will be listened to and addressed.  Without this working group, it may not be obvious who a contributor should reach out to about this type of issue.  Further, it may be easier for the contributor to quit participating all together.

I certainly see no downside to having a group like this.  What are your thoughts on PolicyBrain having a working group for this type of situation?

martinholmer:
@hdoupe asked:
> What are your thoughts on PolicyBrain having a working group for this type of situation?

I don't forsee any problems with having a working group like this.  But things like this are
always benefit-cost tradeoffs.  I can see benefits, and like you, I imagine the cost to be small.
But what if, for some set of reasons, the cost is larger?

hdoupe:
@andersonfrailey do you have any thoughts on the proposed Code of Conduct?

hdoupe:
@GoFroggyRun @brittainhard do you have any thoughts on the proposed Code of Conduct?

andersonfrailey:
@hdoupe I think this is a good start for us. I'd also recommend anyone interested in contributing to the Code of Conduct read through [Astropy's CoC](http://www.astropy.org/code_of_conduct.html). I had the opportunity to hear from one of the coordinators from the project talk about it and she said it has been very helpful for keeping the discussions in their PR's and issues on topic and productive.

MattHJensen:
+1 on the code of conduct. I do wonder whether a dedicated working group adds additional bureaucratic overhead that is really necessary given the current size of the PolicyBrain project.  Could PolicyBrain maintainers serve the same purpose?

GoFroggyRun:
It seems to me that, given the current traffic of PolicyBrain, a separate working group won't be necessary for the moment. When needed, PolicyBrain maintainers could leverage their time and serve the same purpose. In other words, we could just assume PolicyBrain maintainers to be members in the working group.

hdoupe:
@MattHJensen and @GoFroggyRun Perhaps, "working group" is the wrong title.  I just want 2 or 3 people to explicitly be listed as points of contact if someone has a problem.  I am not suggesting that we meet regularly about code of conduct matters.  This group would only meet if there is a problem which seems like what you all are suggesting.

Do you think that for now we should just have a "point of contact group" that includes the core maintainers of this project?

MattHJensen:
I see how a point of contact makes sense. 

Just to give more context to my earlier comments:

I think our aim should be for all project maintainers to share authority for dealing with problems and share responsibility for creating an environment where problems don't happen.

In my view, contributors should also share that responsibility and be able to call each other out for bad behavior (privately or publicly as their judgement dictates) and not feel the need to take everything through a central authority. If something does require a central authority, though, that should be the project maintainers.

hdoupe:
Thanks for your input @MattHJensen, @GoFroggyRun, and @martinholmer.  I tried to incorporate your suggestions.  Is this more in line with what you are thinking?

hdoupe:
@andersonfrailey Thanks for sharing the link to the Astropy CoC.  I really like the tone of the document.  It seems a little more positive than the Pandas CoC.  I'm in favor of replacing the "unacceptable behaviors" section with Astropy's "As members of the community..." section.  What do you think?

andersonfrailey:
@hdoupe I like that. And it seems to be more inline with @MattHJensen's goal of creating an environment where contributors feel comfortable calling out inappropriate comments/behavior without first going through a central authority. When I read "as members of the community..." I interpret it more as how contributors are agreeing to act rather than seeing a list of offenses to be reported.

PR #759, opened 2017-11-30
Quick-calc clean up

### LINKED ISSUE #306: Known issues with "Quick Calculation"
- [ ]  The button does not appear when one chose to the "Edit Parameters" page
- [x] The indexing scheme for how we choose which worker nodes to select for a job is adversely impacted by Quick Calculations. Essentially, the indexing scheme is set up to increment by "NUM_BUDGET_YEARS" but the counting gets off because Quick Calculations are only for one year. The end result is that regular microsim jobs will have their performance adversely impacted.  This will have to be fixed.

### LINKED ISSUE #747: Incorrect caption on "Quick Calculation" button
The caption on the "Quick Calculation" button is no longer correct if TaxBrain is using taxcalc 0.13.2.
Suggest changing it to say "small sample" instead of "2% sample".

screen shot 2017-11-21 at 9 52 40 am

hdoupe:
Resolves #306 and #747


hdoupe:
@MattHJensen This resolves the first item in #306.  Is allowing back-to-back quick-calculations still desired behavior?

cc @martinholmer

hdoupe:
@MattHJensen unless you have any objections, I'm going to merge #759 in an hour or so.

MattHJensen:
No objections. Thanks @hdoupe!

hdoupe:
@MattHJensen Thanks for the review

PR #762, opened 2017-12-01
Save submitted json files in case of warning messages

### LINKED ISSUE #761: warnings on taxbrain/file clear input files
When I upload, for instance, the senate json reform file at taxbrain/file and then click "Show me the results," I get a series of warnings. 

At this point I should be able to click show me the results again to ignore the warnings, but I can't because the reform file I uploaded has been cleared out. 

If I click choose file, reselect the file, and then click run, everything works. 

I shouldn't have to reselect the file. 

This happens on both Chrome and Firefox. Both reform and assumption files are cleared by warnings.

hdoupe:
Resolves #761


hdoupe:
Behavior introduced by #762 

1. If you submit a reform and it throws warnings but not errors, you can simply click the submit calculation button to submit the reform without having to upload the file again.
2. If you submit a reform and it throws warnings but not errors and you upload a new reform file, assumption file, or both new files; your interaction has started over.  Thus, if your new files cause warnings but not errors, then you will have to click the submit calculation button again.
3. As long as your reforms cause errors, you will not be able to submit the file for calculation.

cc @MattHJensen @GoFroggyRun

hdoupe:
PR #762 is on the test app.  I encourage you to play with the new reform file upload logic.  I'd appreciate any feedback.

Some instructions will probably need to be added.  Probably something along the lines of:

1. Upload file and submit reform.
2. If you get warnings but no errors, click submit reform to run the model. If you re-upload a new file (even if it is the exact same one), then you have started the process over.  [Note: we could check to see if an uploaded file is the same as the previous uploaded file. But this seems like an unnecessary head ache.  Why upload the exact same file again?]
3. If you get warnings and errors, you will need to upload a new file.  Once you upload the new file, you have started the process over i.e. if your edited file fixes the errors but causes the same warnings you must still submit the file twice.

Potential file upload button instruction  enhancement:

Currently, the buttons and descriptions look like this:
screen shot 2017-12-07 at 11 12 25 am

Upon submitting the file once and seeing the warnings, should this be changed to:

"Select a new file if you have made changes. Otherwise click 'Submit Reform'"

cc @MattHJensen @GoFroggyRun @martinholmer

martinholmer:
@hdoupe,  Thanks for all the improvement in #762.

If it is easy to do, I'd suggest a small wording change.
If not easy to do, forget the suggestion.

> Currently, the buttons and descriptions look like this:
screen shot 2017-12-07 at 11 12 25 am

How about instead of "Select a reform file" we say "Select a policy reform file"?
A minor change, but it balances the use of an adjective in "Select an economic assumptions file".

hdoupe:
@martinholmer said
> How about instead of "Select a reform file" we say "Select a policy reform file"?
A minor change, but it balances the use of an adjective in "Select an economic assumptions file".

That makes sense to me. I'll make that change.

hdoupe:
I plan to merge #762 this afternoon if there are not any objections.  As a reminder, #762 along with #767 are on the [test app](http://ospc-taxes7.herokuapp.com/taxbrain/file/) right now.

PR #766, opened 2017-12-08
Fix file upload page redirect issue 

### LINKED ISSUE #760: File upload change start year directs to GUI input page
If you select a different start year for the file upload page, then it directs you to the GUI input page with this new start year.  But, if you press the back button, it takes you to the file upload page with the new start year.

The correct behavior should be a redirect to the file upload page with the new start year.

GoFroggyRun:
This PR closes #760. The problem being identified is that correct javascript file is not being properly rendered. In particular, before this PR, some code in `input_base.html` reads:

```
  {% if input_type == "file" %}
  <script src="{% static 'js/filetb.js' %}"></script>
  {% else %}
  <script src="{% static 'js/taxbrain.js' %}"></script>
  {% endif %}
```

The if loop here, however, will always fall into the `else` clause, and thus causing #760. 

An easy yet inelegant solution would be having two separate templates for `input_form.html` and `input_file.html`, which is what I used in this PR. 

While working out more concise and elegant solution, it'd also be a good idea to test this PR on test server. 

@hdoupe @MattHJensen

### DIFF
--- a/templates/taxbrain/file_base.html
+++ b/templates/taxbrain/file_base.html
@@ -86,7 +86,7 @@
   <script src="{% static 'js/vendor/share-button/share-button.min.js' %}"></script>
   <!-- Scripts -->
   <script src="{% static 'js/site.js' %}"></script>
-  <script src="{% static 'js/taxbrain.js' %}"></script>
+  <script src="{% static 'js/filetb.js' %}"></script>
   <script src="{% static 'js/forms.js' %}"></script>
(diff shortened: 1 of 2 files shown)


hdoupe:
@GoFroggyRun I think I found the root of the problem.  I think #760 should be closed in favor of #761.  What do you think?

GoFroggyRun:
Closing in favor of #767

Issue #773, opened 2017-12-11
Python version and other environment questions


hdoupe:
While resetting my local environment, I got the following error:
```
Install webapp-public requirements.txt
Collecting Django==1.8.2 (from -r ../requirements.txt (line 1))
/anaconda/envs/aei_dropq/lib/python2.7/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:318: SNIMissingWarning: An HTTPS request has been made, but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may cause the server to present an incorrect TLS certificate, which can cause validation failures. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.
  SNIMissingWarning
/anaconda/envs/aei_dropq/lib/python2.7/site-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a newer version of Python to solve this. For more information, see https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
  Could not fetch URL https://pypi.python.org/simple/django/: There was a problem confirming the ssl certificate: [Errno 1] _ssl.c:504: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed - skipping
  Could not find a version that satisfies the requirement Django==1.8.2 (from -r ../requirements.txt (line 1)) (from versions: )
No matching distribution found for Django==1.8.2 (from -r ../requirements.txt (line 1))
FAILED
~/Documents/PolicyBrain
```

This occurred when setting requirements in the [`requirements.txt`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/7da856c1740023ab245bf46e5a1863d60748da10/requirements.txt) file from this function in [`install_taxbrain_server.sh`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/deploy/install_taxbrain_server.sh#L19-L20).

[...]

PR #776, opened 2017-12-12
Short-term fix for package pinning

hdoupe:
This PR provides a short-term fix for the environment issues discussed in #773.

### DIFF
--- a/deploy/fab/dropq_environment.yml
+++ b/deploy/fab/dropq_environment.yml
@@ -1,6 +1,6 @@
 name: aei_dropq
 dependencies:
-- python=2.7.12
+- python>=2.7.12
 - pandas>=0.20.1
 - flask
 - greenlet


martinholmer:
@hdoupe,  Is there are reason to **exact pin** the `eventlet` package?
```
   - eventlet==0.20.0
```

hdoupe:
@martinholmer my approach to environment setup so far has been to change as little as possible since I still do not understand why some packages are exact pins and some are not.  This is my first time getting into the weeds of how this part works.  In the next few months when there are less time sensitive demands, I plan to clean this up.

martinholmer:
@hdoupe said:
> my approach to environment setup so far has been to change as little as possible since I still do not understand why some packages are exact pins and some are not. This is my first time getting into the weeds of how this part works. In the next few months when there are less time sensitive demands, I plan to clean this up.

Sounds quite sensible.

PR #777, opened 2017-12-13
Update warning/error messages and description

### LINKED ISSUE #769: Warning message clarifications
In open-source-economics/Tax-Calculator#1736, @martinholmer said

> The warning messages on TaxBrain are so garbled that they are uninformative, and while it does say this, it is not clear enough that these are just warnings (not errors) and that clicking on the "Show Me the Results" button a second time will submit the run for execution. I'm pretty sure that the TaxBrain developers were working this week on these problems with the TaxBrain file upload page. Hopefully the next version of TaxBrain will be more clear. Meanwhile, see if clicking a second time on the "Show Me the Results" button produces results for the reform.

It seems like there are several issues here:
1. It is not clear that if you click "Show Me the Results" button a second time that the model will run.   I thought that the first message makes this point pretty clearly.  However, if this is not the case, then I am open to suggestions for improving the presentation of this message.
screen shot 2017-12-10 at 1 01 00 pm

2. "The warning messages on TaxBrain are so garbled that they are uninformative."  The current warning/error message logic was first implemented in #641.  We decided in favor of verbose warning and error messages.  In this [comment](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/641#issuecomment-329578162), I suggested some different approaches that we could revisit.  Development time is pretty tight right now.  So, I cannot promise an immediate update of the warning/error display logic.  However, I am open to discussing ideas.  Further, if someone wants to open a PR to change this, then the relevant function is [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/641#issuecomment-329578162).  

3. The distinction between warning and error messages is not obvious.  As @martinholmer suggested, perhaps the warning messages should be yellow and the error messages should be red.  This makes sense to me as a long-term solution.  In the meantime, we can try to make the initial message more informative.  Currently, it says:
"Some fields have errors. Values outside of suggested ranges will be accepted if they only cause warnings and are submitted again from this page."
I think we could change it to say:
"Some fields have warnings or errors. Values outside of suggested ranges will be accepted if they only cause warnings and are submitted again from this page.  Warning messages begin with "WARNING", and error messages begin with "ERROR"."

4. Further specification needs to be made for the warning/error messages.  Currently, the message does not include the parameter name.  This is an easy fix and should be updated immediately.

cc @martinholmer @MaxGhenis

hdoupe:
This PR begins to resolve #769 part 3 and resolves #769 part 4.

1. Adds a more informative description for what a warning and error message look like:
screen shot 2017-12-12 at 9 08 30 pm

2. Includes offending parameter in the message.  This was not needed in the GUI messages since it was obvious that the message applied to the parameter whose box it was displayed under.  As @martinholmer pointed out, it is not so obvious when the messages are displayed on the file input page.
screen shot 2017-12-12 at 9 10 41 pm

### DIFF
--- a/webapp/apps/taxbrain/tests/test_views.py
+++ b/webapp/apps/taxbrain/tests/test_views.py
@@ -565,7 +565,7 @@ class TaxBrainViewsTests(TestCase):
         # Check that no redirect happens
         self.assertEqual(response.status_code, 200)
         assert response.context['has_errors'] is True
-        msg = 'ERROR: value 9e+99 > max value 89239.88 for _II_brk2_4 for 2024'
+        msg = 'ERROR: _II_brk1_4 value 9e+99 > max value 89239.88 for _II_brk2_4 for 2024'
         assert msg in response.context['errors']
 
         # get most recent object
@@ -601,7 +601,7 @@ class TaxBrainViewsTests(TestCase):
         # Check that no redirect happens
         self.assertEqual(response.status_code, 200)
         assert response.context['has_errors'] is True
-        msg = 'WARNING: value 1073.53 < min value 7191.08 for 2023'
+        msg = 'WARNING: _STD_0 value 1073.53 < min value 7191.08 for 2023'
         assert msg in response.context['errors']
 
         # get most recent object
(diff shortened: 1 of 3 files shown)


hdoupe:
PR #777 is on the [test app](http://ospc-taxes7.herokuapp.com/taxbrain/) along with PolicyBrain 1.2.1.

PR #780, opened 2017-12-18
Fix buttons disappear after fiscal tables being re-rendered

### LINKED ISSUE #613: Revenue table (Print, etc.) buttons disappear when selecting different (Reform, etc.) table
@GoFroggyRun after looking at this issue, it looks like the buttons for the table aren't being re-rendered when a different table is looked at. Unfortunately, the buttons are added after the table has already been rendered. 

The buttons are added here: 
https://github.com/OpenSourcePolicyCenter/webapp-public/blob/master/static/js/taxbrain-tablebuilder.js#L554-L567

But then need to be added again here after re-rendering: 
https://github.com/OpenSourcePolicyCenter/webapp-public/blob/master/static/js/taxbrain-tablebuilder.js#L261-L272

So, the fiscal table is being re-rendered but it's getting rid of the buttons each time it's being re-rendered. Hope that makes sense.

I spent a couple hours looking at it and couldn't make headway. I can take another look tomorrow.

GoFroggyRun:
This PR closes #613

### DIFF
--- a/static/js/taxbrain-tablebuilder.js
+++ b/static/js/taxbrain-tablebuilder.js
@@ -262,7 +262,6 @@ $(function() {
                 e.preventDefault();
                 var $el = $(e.currentTarget);
                 var el_tablename = $el.data('tablename');
-
                 if (this.selectedTableName === el_tablename) {
                   return;
                 }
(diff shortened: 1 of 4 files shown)


GoFroggyRun:
@hdoupe can we have this PR included in the new update?

hdoupe:
Looks good to me. Thanks @GoFroggyRun

Issue #763, opened 2017-12-01
Issues with 'cpi_offset' parameter


GoFroggyRun:
Both the [House](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_House_Amended.json) and the [Senate](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/reforms/TCJA_Senate_111417.json) adopt chained-CPI indexing. While working on the reform file upload option and TB's GUI input work, I realized that there might be some issues with how this parameter works.

To realize chained-CPI indexing for year **2018**, the `cpi_offset` parameter needs to be set to `-0.0025` in year **2017**. However, both reforms (House and Senate) actually take affect in **2018**. in this case, to implement either reform in TB's GUI, users have to **always** begin with year **2017**. So each input field other than `cpi_offset` has to have an extra `*`, indicating no change in 2017's law. Also, the result table would build an extra column for year 2017 with no change whatsoever. Does this really make sense?

A couple solutions to simplify the issue: 

1. Come up with some ways to allow users to edit previous year parameter on TB's GUI page, just like what we have done to allow future year edits.

2. Modify how `cpi_offset` parameter work in tax-calculator by skimpily implementing the chained-CPI indexing one year head of reform year specified.

@martinholmer @MattHJensen @hdoupe

martinholmer:
@GoFroggyRun said we could:
> Modify how `cpi_offset` parameter work in Tax-Calculator by skimpily implementing the chained-CPI indexing one year head of reform year specified.

I view this as a very bad idea and I'm not even sure how it would work.  Tax-Calculator has no problem with a reform that starts before the ten-year budget window.  It is the design of TaxBrain that is the problem.  When I started on this project, TaxBrain required all reform provision to start in the first budget year (that is, there was no `*` capability to specify delayed implementation of a reform provision).

martinholmer:
@GoFroggyRun said we could:
> Come up with some ways to allow users to edit previous year parameter on TaxBrain's GUI page, just like what we have done to allow future year edits.

And @martinholmer said:
> Tax-Calculator has no problem with a reform that starts before the ten-year budget window. It is the design of TaxBrain that is the problem.

[...]

PR #782, opened 2017-12-19
Convert DependentCredit_before_CTC from bool to int

### LINKED ISSUE #765: Handle future year reforms for checkboxes
There's one provision in TCJA Senate Modified Mark's proposal that introduces "stack[ing] the dependent credit before the child tax credit" in 2018, and then repeals it starting 2026. 

Currently, this boolean parameter `_DependentCredit_before_CTC` is handled using a checkbox on TaxBrain: checked box indicates true while unchecked one represents false, which I believe is a concise way of integrating such boolean parameters. It seems that, however, such format won't allow users to specify any further reforms for future years.  

Since it would be desirable to handle both House and Senate reforms in TaxBain's GUI, we might want to modify, at least for `_DependentCredit_before_CTC` parameter, the way of handling boolean fields to allow future year reforms. However, I haven't come up with an ideal way to deal with it. 

Any comments, concerns or remarks would be helpful. @martinholmer @MattHJensen @hdoupe

hdoupe:
Resolves #765.  This PR treats `DependentCredit_before_CTC` as an integer instead of a boolean parameter which allows the user to very the parameter's value over time.

EDIT:
Old behavior:
screen shot 2017-12-19 at 5 12 11 pm

New behavior:
screen shot 2017-12-19 at 5 12 49 pm

### DIFF
--- a/webapp/apps/btax/forms.py
+++ b/webapp/apps/btax/forms.py
@@ -3,12 +3,12 @@ from django.forms import ModelForm
 from django.utils.translation import ugettext_lazy as _
 
 from .models import BTaxSaveInputs
-from .helpers import (BTaxField, BTaxParam, get_btax_defaults)
+from .helpers import (BTaxField, BTaxParam, get_btax_defaults,
+                      convert_val, make_bool)
 from ..taxbrain.helpers import (is_number, int_to_nth,
                                 is_string, string_to_float_array,
                                 check_wildcards, expand_list,
-                                propagate_user_list, convert_val,
-                                make_bool)
+                                propagate_user_list)
 import taxcalc
 
 from ..taxbrain.forms import (has_field_errors,
--- a/webapp/apps/btax/helpers.py
+++ b/webapp/apps/btax/helpers.py
@@ -210,3 +210,17 @@ def group_args_to_btax_depr(btax_default_params, asset_yr_str):
 
             )
     return depr_argument_groups
+
+
+def make_bool(x):
+    b = True if x == 'True' else False
+    return b
+
+
+def convert_val(x):
+    if is_wildcard(x):
+        return x
+    try:
+        return float(x)
+    except ValueError:
+        return make_bool(x)
--- a/webapp/apps/btax/views.py
+++ b/webapp/apps/btax/views.py
@@ -38,11 +38,10 @@ from .models import BTaxSaveInputs, BTaxOutputUrl
 from .helpers import (get_btax_defaults,
                       BTAX_BITR, BTAX_DEPREC,
                       BTAX_OTHER, BTAX_ECON,
-                      group_args_to_btax_depr, hover_args_to_btax_depr)
+                      group_args_to_btax_depr, hover_args_to_btax_depr,
+                      make_bool, convert_val)
 from ..taxbrain.helpers import (format_csv,
-                                is_wildcard,
-                                convert_val,
-                                make_bool)
+                                is_wildcard)
 from ..taxbrain.views import (benefit_switch_fixup,
                               denormalize, normalize)
 from .compute import DropqComputeBtax, MockComputeBtax, JobFailError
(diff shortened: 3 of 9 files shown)


hdoupe:
Are there any other boolean parameters that we would like to convert into comma separated value parameters?  

My plan is to only convert the ones where we see an obvious use case for allowing changes over time.  If we like the behavior introduced for these select parameters, then we can convert the rest of the boolean parameters.  I think this will minimize the number of changes to the database.

cc @MattHJensen @martinholmer @codykallen

MattHJensen:
@hdoupe, once you feel comfortable with the database changes, I think we should make this change for all boolean parameters, as it is very difficult to anticipate what could or couldn't be phased in over time or sunset.

martinholmer:
@hdoupe said:
> This [PolicyBrain] PR [#782] treats `DependentCredit_before_CTC` as an integer instead of a boolean parameter which allows the user to very the parameter's value over time.

OK, but this parameter is boolean in Tax-Calculator.  So that means the JSON reform file TaxBrain generates must have them as boolean even though they are presented as 0s and 1s in the TaxBrain GUI.

Is there a reason the GUI can't use comma-separated boolean values?

hdoupe:
@MattHJensen Ok, that makes sense.  I'll get this on the test app today.  If everything goes well, then I will apply this change to the rest of the boolean parameters.

hdoupe:
@martinholmer asked:
> OK, but this parameter is boolean in Tax-Calculator. So that means the JSON reform file TaxBrain generates must have them as boolean even though they are presented as 0s and 1s in the TaxBrain GUI.
> 
> Is there a reason the GUI can't use comma-separated boolean values?

The GUI can tolerate comma separated boolean values.  However, I think that we should also allow the submission of 1/0 since they are logically equivalent.

hdoupe:
@martinholmer I'm getting the following error when I submit this reform:
screen shot 2017-12-20 at 10 05 15 am

A simpler example of this reform is run here:
```
import taxcalc

policy_dict = {
    u'policy': {
        2019: {
            u'_DependentCredit_before_CTC': [1],
        },
    },
    u'growdiff_response': {},
    u'consumption': {},
    u'behavior': {},
    u'growdiff_baseline': {}
}

ew  = taxcalc.tbi.reform_warnings_errors(policy_dict)

print("errors warnings for: ")
print(policy_dict)
print('\n')
print(ew)

policy_dict["policy"][2019]["_DependentCredit_before_CTC"] = [1.0]

ew = taxcalc.tbi.reform_warnings_errors(policy_dict)
print("\nerrors warnings for: ")
print(policy_dict)
print('\n')
print(ew)
```

I get the following output:
```
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ python tc_pr_1775.py 
errors warnings for: 
{u'policy': {2019: {u'_DependentCredit_before_CTC': [1]}}, u'growdiff_response': {}, u'behavior': {}, u'consumption': {}, u'growdiff_baseline': {}}


{'errors': '', 'warnings': ''}

errors warnings for: 
{u'policy': {2019: {u'_DependentCredit_before_CTC': [1.0]}}, u'growdiff_response': {}, u'behavior': {}, u'consumption': {}, u'growdiff_baseline': {}}


{'errors': 'ERROR: 2019 _DependentCredit_before_CTC value 1.0 is not boolean\n', 'warnings': ''}
```

It seems like 1.0 is not a valid boolean according to Tax-Calculator even though we also have:
```
In [1]: bool(1.0) == bool(1) == bool(True)
Out[1]: True
```

PolicyBrain reads the user input from the GUI as strings.  The strings with digits are converted to python floats.  I could add a special case for the boolean parameters that are entered as 1/0/1.0/0.0 that converts them to 1/0 or True/False.  However, I think a better solution would be for Tax-Calculator to accept 1.0 and 0.0 as valid boolean values.  @martinholmer what are your thoughts on this?

MattHJensen:
Just my 2 cents, I am more eager for TaxBrain to be able to accept True/False than for either TaxBrain or Tax-Calculator to accept 1.0, 1, 0, 0.0. We just have so many numeric inputs, including switches, it refreshing to differentiate switches.

hdoupe:
@MattHJensen Thanks for your insight.  I moved your comment to issue #786.  Sorry, I probably should have left this PR open longer.  I was in a bit of a rush to get this merged, get a pre-release up, and move on to #763.  I should have waited longer to make sure that we were all on board.

PR #785, opened 2017-12-20
Update to PolicyBrain 1.2.2 --> 1.3.0

hdoupe:
Updates RELEASES.md and hardcoded version for PolicyBrain 1.2.2.


hdoupe:
I think it's a good idea to do a bump in the minor version with the addition of PR's #788 and #790 

@MattHJensen @GoFroggyRun Does this sound sensible?

hdoupe:
@MattHJensen and @GoFroggyRun we are close to finishing this 1.3.0.  I'd like to merge #790, #791, and #792; do another pre-release and push that to the test app; and do a final release if that release candidate checks out.

Would you mind taking a final look at #790, #791 and #792?

martinholmer:
#792 fails the test suite and is wrong in one place as I said in #792

PR #788, opened 2017-12-21
Convert default values for boolean parameters from 1/0 to "True"/"False"

hdoupe:
This PR converts 1/0 values back to boolean variables and serializes them.  PolicyBrain reads the default GUI parameter values from Tax-Calculator's [`default_data`](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/parameters.py#L24) function.  This [line in parameters.py](https://github.com/open-source-economics/Tax-Calculator/blob/master/taxcalc/parameters.py#L438) converts boolean parameters from true/false to 1/0.  This treatment is currently only applied to `DependentCredit_before_CTC` but will be applied to the other comma separated boolean variables as they are added.

Before this PR, the default parameter value for `DependentCredit_before_CTC` was shown as:
screen shot 2017-12-21 at 5 52 17 pm

Behavior introduced by this PR shows:
screen shot 2017-12-21 at 5 53 14 pm

### DIFF
--- a/webapp/apps/taxbrain/helpers.py
+++ b/webapp/apps/taxbrain/helpers.py
@@ -809,6 +809,18 @@ class TaxCalcParam(object):
         # organize defaults by column [[A1,B1],[A2,B2]] to [[A1,A2],[B1,B2]]
         values_by_col = [list(x) for x in zip(*values_by_year)]
 
+        # Tax-Calculator converts boolean values to 1/0 via
+        # np.array(bool_val, np.int8)
+        # here we convert that value back to a boolean type and serialize it
+        if self.nice_id in BOOL_PARAMS:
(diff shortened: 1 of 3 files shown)


hdoupe:
PR #788 is on the test-app.  I hope to merge this and push to production tomorrow if there are no objections.

MattHJensen:
@hdoupe, I just tried entering False, False, False and got an error. Same with True, True, True

`Expected case insensitive 'true' or 'false' but got  True`

hdoupe:
Thanks for the bug report @MattHJensen.  I just updated the test app with the bug fix.

MattHJensen:
@hdoupe, everything is looking good to me!

PR #797, opened 2018-01-09
Remove Expanded-Income/or/Adjusted-Income row buttons

### LINKED ISSUE #796: Need to remove Expanded-Income/or/Adjusted-Income row of buttons
Now that we have decided that TaxBrain will never show distribution tables or difference tables with the AGI as the classifier variable, the expanded-income/or/adjusted-income row of buttons needs to be removed.  Here is what thinks look like in PolicyBrain 1.3.0:

screen shot 2018-01-09 at 11 34 12 am

That third row of buttons needs to be removed.
If there is some delay in removing them, then the tool tip should be changed from "Not yet available" to "Not a choice" or "Never will be available" or something less misleading than the current text.

GoFroggyRun:
Per @martinholmer's suggestion, this PR closes #796. The build-table option would look like the following after this PR:

screen shot 2018-01-09 at 12 41 09 pm

Corresponded tooltip constants have been removed as well. 

cc @hdoupe


martinholmer:
@GoFroggyRun,  Thanks for the quick resolution to issue #796 in pull request #797.

hdoupe:
@GoFroggyRun is #797 ready for review?

GoFroggyRun:
@hdoupe Yup.

hdoupe:
Looks good to me. Thanks @GoFroggyRun

PR #800, opened 2018-01-11
Refactor warning/error parsing

hdoupe:
This PR refactors the warning/error parsing logic. This reduces duplicated code and puts the warning/error parsing logic under test.

cc @codekansas

### DIFF
--- a/webapp/apps/taxbrain/views.py
+++ b/webapp/apps/taxbrain/views.py
@@ -6,6 +6,7 @@ import pytz
 import os
 import tempfile
 import re
+from collections import defaultdict
 
 #Mock some module for imports because we can't fit them on Heroku slugs
 from mock import Mock
@@ -175,7 +176,7 @@ def parse_errors_warnings(errors_warnings, map_back_to_tb):
     returns: dictionary 'parsed' with keys: 'errors' and 'warnings'
         parsed['errors/warnings'] = {year: {tb_param_name: 'error message'}}
     """
-    parsed = {'errors': {}, 'warnings': {}}
+    parsed = {'errors': defaultdict(dict), 'warnings': defaultdict(dict)}
     for action in errors_warnings:
         msgs = errors_warnings[action]
         if len(msgs) == 0:
@@ -188,8 +189,6 @@ def parse_errors_warnings(errors_warnings, map_back_to_tb):
             year = msg_spl[1]
             curr_id = msg_spl[2][1:]
             msg_parse = msg_spl[2:]
-            if curr_id not in parsed[action]:
-                parsed[action][curr_id] = {}
             parsed[action][curr_id][year] = ' '.join([msg_action] + msg_parse +
                                                          ['for', year])
 
--- a/webapp/apps/taxbrain/views.py
+++ b/webapp/apps/taxbrain/views.py
@@ -171,11 +171,11 @@ def parse_errors_warnings(errors_warnings, map_back_to_tb):
             msg_spl = msg.split()
             msg_action = msg_spl[0]
             year = msg_spl[1]
-            curr_id = msg_spl[2]
+            curr_id = msg_spl[2][1:]
             msg_parse = msg_spl[2:]
-            if year not in parsed[action]:
-                parsed[action][year] = {}
-            parsed[action][year][curr_id[1:]] = ' '.join([msg_action] + msg_parse +
+            if curr_id not in parsed[action]:
+                parsed[action][curr_id] = {}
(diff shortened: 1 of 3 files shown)


webapp/apps/test_assets/test_param_formatters.py
```diff
@@ -3,7 +3,10 @@
 import taxcalc
 import numpy as np
 
-from ..taxbrain.views import read_json_reform, parse_errors_warnings
+from ..taxbrain.views import (read_json_reform, parse_errors_warnings,
```
codekansas: One handy software engineering tool you might be interested in is "blame". Basically, it's a tool that shows the last person to make changes to each line of code. GitHub has a more thorough explanation [here](https://help.github.com/articles/tracing-changes-in-a-file/). This isn't a review comment per se, but one thing to keep in mind when using that tool is writing code in a way that "preserves blame", in other words, so that you don't make changes to lines that you don't edit. In this case, the best way to "preserve blame" would be to write each import on a new line, so that if you add your own import, it doesn't update the blame on someone else's import, so you know the last person who thought that import was important to change. i.e.

```python
from ..taxbrain.views import (
   read_json_reform,
   parse_errors_warnings,
   ...
)
```

webapp/apps/test_assets/test_reform.py
```diff
@@ -372,30 +372,34 @@
...
+        'STD_3': {
+            '2020': 'WARNING: _STD_3 value 150.0 < min value 9900.32 for 2020',
+            '2021': 'WARNING: _STD_3 value 153.57 < min value 10138.33 for 2021',
+            '2022': 'WARNING: _STD_3 value 157.3 < min value 10387.12 for 2022',
+            '2023': 'WARNING: _STD_3 value 161.03 < min value 10635.66 for 2023',
+            '2024': 'WARNING: _STD_3 value 164.89 < min value 10893.32 for 2024',
+            '2025': 'WARNING: _STD_3 value 168.81 < min value 11154.96 for 2025',
+            '2026': 'WARNING: _STD_3 value 172.83 < min value 11422.83 for 2026',
+            '2027': 'WARNING: _STD_3 value 176.98 < min value 11699.38 for 2027'
+        }
```
codekansas: One software engineering practice you might want to consider is separating "code" from "data". In this case the "data" is these warnings, which could be saved as like a separate JSON file somewhere and loaded into Python (using the default `import json` library). The main goal is to abstract away this kind stuff from the core code logic, which makes your code much more readable.

webapp/apps/taxbrain/views.py
```diff
@@ -151,6 +151,36 @@ def normalize(x):
     return ans
 
 
+def append_ew_file_input(error_list, msg, *args):
```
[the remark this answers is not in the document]
hdoupe: I don't have any aversion to lambdas. I think this looks a little cleaner.

Out of curiosity, why do some people frown upon lambda functions?

webapp/apps/test_assets/test_param_formatters.py
hdoupe: I hadn't thought about preserving "blame" before. That makes a lot of sense. Thanks for the pointer.

webapp/apps/test_assets/test_reform.py
hdoupe: Ok, I see. I'll open a follow up PR to clean up `test_reform.py`. There are plenty of data there that need to be moved into JSON files.

webapp/apps/taxbrain/views.py
```diff
@@ -171,11 +201,11 @@ def parse_errors_warnings(errors_warnings, map_back_to_tb):
             msg_spl = msg.split()
             msg_action = msg_spl[0]
             year = msg_spl[1]
-            curr_id = msg_spl[2]
+            curr_id = msg_spl[2][1:]
             msg_parse = msg_spl[2:]
-            if year not in parsed[action]:
-                parsed[action][year] = {}
```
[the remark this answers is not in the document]
hdoupe: Oh nice. `defaultdict` is pretty nifty. I wish I'd known about `defaultdict` earlier.

hdoupe:
@codekansas Thanks for the review. Let me know what you think of the changes.

webapp/apps/test_assets/test_param_formatters.py
codekansas: Check out the blame tool, it is useful for figuring out who wrote something if you're trying to figure out why it was done that way (i.e. poorly documented code)

codekansas:
@hdoupe LGTM

hdoupe:
Thanks for the review @codekansas

PR #810, opened 2018-01-22
Small changes to .gitignore and install script

codekansas:
Explanation:

- `.DS_Store` - Useless file created by Mac file browser
- `.egg-info` - Created when doing pip install
- `staticfiles` - Migrated staticfiles, should probably be ignored
- `pip install -e .` - Safer / more consistent version of `python setup.py develop`

I talked to @hdoupe about my regular schedule now so expect commits around this time MWF/Sat/Sun.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,8 @@ dump.rdb
 *.pem
 
 _build
+
+.DS_Store
+*.egg-info
+staticfiles
+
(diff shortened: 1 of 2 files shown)


hdoupe:
Thanks for the contribution @codekansas. We've always kept the `staticfiles` in the repo, and I was never really sure why. @GoFroggyRun is there a back-story for this? I just want to make sure that they weren't added as some Heroku work around or something.

GoFroggyRun:
@hdoupe I'm not sure about `staticfiles` either: they were there before I worked on Policy Brain repo. @martinholmer had similar concern before, but we didn't get any answer from Continuum.

hdoupe:
@GoFroggyRun Ok, thanks. I think that we should just remove them. If there's a problem for some reason, then we can add them back.

codekansas:
@hdoupe ping

hdoupe:
Thanks for the contribution @codekansas

Issue #811, opened 2018-01-24
Save model parameters as JSON objects


hdoupe:
In PR #808, we discovered a bug involving how model parameters are saved. That sparked the following conversation:

@hdoupe said
> Fixing this under the current approach of storing each parameter to a column in a database would involve extensive changes to the database. Given the magnitude of these changes, I think this is a good time to make the swap to storing the run parameters primarily as JSON files. This idea was first raised in #435, @GoFroggyRun has mentioned this, and @codekansas has also suggested that we move away from our current approach.
> 
> This would have no effect on the file upload runs, and the raw input from the GUI is parsed as a dictionary that is built from the attributes of the model object (i.e. taxbrain_data = dict(taxbrain_model.__dict__)). These attributes are populated by the raw input data. Thus, it would be easy to make the change going forward, but it may be more difficult to preserve backwards compatibility.
> 
> One solution is to create a decorator like we did for the new results table names. If the version of PB is less than 1.4.0 (would be a minor bump to 1.4.0 if we do this change), then we form the JSON object from the raw fields and serve that to the parsing functions. This would require saving all of the no longer needed fields though.
> 
> @GoFroggyRun @codekansas what do you think?

@GoFroggyRun said:
> @hdoupe said:
> > I think this is a good time to make the swap to storing the run parameters primarily as JSON files.
> 
> Definitely agree.
> 
> Regarding backward compatibility, I'm not sure whether we want to store "all of the no longer needed fields." Would it be easier to preserve backward compatibility if we just provide the reform file ran, not necessarily via GUI, and its results (which I think is the point of preserving the compatibility)? We don't have to worry too much when fields were added or removed, while users are able to archive what they ran along with the results and cite whatever they need.

@hdoupe said:
> Right, they could pull up the results just fine no matter what we do with the parameters. However, we need the fields data to provide the edit parameters page for previous runs. I think there are a couple solutions:
> 
> 1. JSON file to fields converter--as long as we have the JSON reform (which we do from PB 1.0.2 onward) we can convert that back to the fields data and display that. I've been interested in having something like this for awhile, but it may take some time to get right. And there may be some edge cases that are difficult to catch.
> 2. One time database update that converts all stored run parameters into JSON fields objects and store those. Then delete the old, no-longer needed columns.
> 3. Keep the old columns and use some decorator attribute to dynamically create the JSON fields object and parse that.
> 
[...]

PR #814, opened 2018-01-25
Resolve more backwards compatibility issues related to old taxcalc tables

### LINKED ISSUE #812: Accidentally deleted tax_result column
I accidentally deleted the tax_result column in the TaxBrain database for runs occurring before 11/22/2017. Heroku only saves database history for the previous week. Fortunately, I pulled the production and test databases down on 11/09/2017 and still have them on my laptop. So, I should be able to save the old tax runs. However, there was a period from 11/09/2017 to 11/22/2017 for which the tax results are gone.

When I created a decorator for the `tax_result` attribute in PR #738  I added a field `_tax_result` to the Django model. However, when I did the migration, instead of changing the name from `tax_result` to `_tax_result`, I removed `tax_result` and added a new field `_tax_result`.

### LINKED ISSUE #813: Casting Error in version check
Run 1400 on the test app throws the following error:
screen shot 2018-01-25 at 3 27 35 pm

In older versions of PolicyBrain, the package version was stored as `0.9.0.d79abf` for example. thus, the following [snippet](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/webapp/apps/taxbrain/models.py#L772-L785) of code throws a value error:
```
taxcalc_vers = tuple(map(int, taxcalc_vers.split('.')))
```

hdoupe:
Resolves #813 

[EDIT] Also resolves #812


webapp/apps/dynamic/models.py
```diff
@@ -87,6 +87,25 @@ class DynamicBehaviorSaveInputs(models.Model):
...
+        # older PB versions stored commit reference too
+        # e.g. taxcalc_vers = "0.9.0.d79abf"
+        if len(taxcalc_vers) >=3:
+            taxcalc_vers = taxcalc_vers[:3]
+        taxcalc_vers = tuple(map(int, taxcalc_vers))
+        if taxcalc_vers >= (0, 13, 0):
+            return self.tax_result
+        else:
+            return rename_keys(self.tax_result, PRE_TC_0130_RES_MAP)
+
```
hdoupe: @codekansas I would prefer to not copy this method from the taxbrain models code. Do you have any ideas for a cleaner solution?

hdoupe:
#812 was caused because I added a decorator for the `tax_result` variable. In doing so, I added a field `_tax_result` that pointed to a column in the `taxbrain_taxsaveinputs` table, `tax_result`. However, in the migrations file, you can see that the field `tax_result` was [removed from the table](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/9edd4a52626c99956b863ac1a3634b8e9bd31b40/webapp/apps/taxbrain/migrations/0056_auto_20171120_2123.py#L16-L19) and then, [added back later](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/9edd4a52626c99956b863ac1a3634b8e9bd31b40/webapp/apps/taxbrain/migrations/0056_auto_20171120_2123.py#L180-L184), effectively deleting all previous information.

This PR uses a method to do the table formatting instead of a decorator. I'd be interested in hearing feedback on this approach and ideas for other approaches.

webapp/apps/dynamic/models.py
```diff
@@ -87,6 +87,25 @@ class DynamicBehaviorSaveInputs(models.Model):
...
 
+    def get_tax_result(self):
+        """
+        If taxcalc version is greater than or equal to 0.13.0, return table
+        If taxcalc version is less than 0.13.0, then rename keys to new names
+        and then return table
+        """
+        outputurl = OutputUrl.objects.get(unique_inputs__pk=self.pk)
+        taxcalc_vers = outputurl.taxcalc_vers
+        taxcalc_vers = taxcalc_vers.split('.')
```
codekansas: Can simplify to

```
taxcalc_vers = outputurl.taxcalc_vers.split('.')
```

But probably the best solution would be to use some versioning comparison standard library, i.e. `distutils.version.StrictVersion` or `distutils.version.LooseVersion` so that you don't have to mess around with this version comparison stuff.

webapp/apps/dynamic/models.py
codekansas: You'll probably want to put this in a `utils.py` file in the parent directory. Hmm... If you do this, you'll probably also want to include

```
from __future__ import absolute_import
```

in the top of the file, so that importing is consistent.

webapp/apps/taxbrain/models.py
```diff
@@ -768,25 +768,24 @@ class TaxSaveInputs(models.Model):
     # Creation DateTime
     creation_date = models.DateTimeField(default=datetime.datetime(2015, 1, 1))
 
-
-    @property
-    def tax_result(self):
+    def get_tax_result(self):
```
codekansas: This seems like a good idea - I think the best "logic" for deciding if something should be a property is if:

1. It is computationally light, or
2. It is cached (i.e. the result is only computed once)

webapp/apps/taxbrain/models.py
```diff
@@ -768,25 +768,24 @@ class TaxSaveInputs(models.Model):
...
         """
         If taxcalc version is greater than or equal to 0.13.0, return table
         If taxcalc version is less than 0.13.0, then rename keys to new names
         and then return table
         """
         outputurl = OutputUrl.objects.get(unique_inputs__pk=self.pk)
         taxcalc_vers = outputurl.taxcalc_vers
-        taxcalc_vers = tuple(map(int, taxcalc_vers.split('.')))
+        taxcalc_vers = taxcalc_vers.split('.')
```
codekansas: See earlier comment about using `distutils` for version parsing instead

webapp/apps/taxbrain/views.py
```diff
@@ -1014,7 +1014,7 @@ def csv_output(request, pk):
     filename = "taxbrain_outputs_" + suffix + ".csv"
     response['Content-Disposition'] = 'attachment; filename="' + filename + '"'
```
codekansas: It is probably better to use the following:

```
filename = 'taxbrain_outputs_{}.csv'.format(suffix)
response['Content-Disposition'] = 'attachment; filename="{}"'.format(filename)
```

Also: For debugging purposes, it is a good idea to [log things](https://docs.djangoproject.com/en/2.0/topics/logging/) (the response filename, for example).

webapp/apps/dynamic/models.py
hdoupe: @codekansas thanks for the advice. That makes sense.

Another solution would be to create an abstract model class with this method. Then, the `TaxSaveInputs` and `DynamicSaveInputs` classes could inherit this class. It is likely that we will have to add a `get_fields` method in a separate PR that resolves #811. This would be like the base class approach described [here](https://medium.com/eshares-blog/supercharging-django-productivity-at-eshares-8dbf9042825e). An example that is similar to our situation is described [here](http://blog.kevinastone.com/django-model-behaviors.html).

webapp/apps/taxbrain/models.py
hdoupe: Ok, that makes sense. Thanks for the info. I also couldn't get the property approach to work without deleting the column in the database.

webapp/apps/dynamic/models.py
```diff
@@ -87,6 +87,27 @@ class DynamicBehaviorSaveInputs(models.Model):
...
+        """
+        If taxcalc version is greater than or equal to 0.13.0, return table
+        If taxcalc version is less than 0.13.0, then rename keys to new names
+        and then return table
+        """
+        outputurl = OutputUrl.objects.get(unique_inputs__pk=self.pk)
+        taxcalc_vers = outputurl.taxcalc_vers
+        # only the older (pre 0.13.0) taxcalc versions are null
+        if taxcalc_vers:
+            taxcalc_vers = LooseVersion(taxcalc_vers)
```
codekansas: I feel like this should break something because LooseVersion isn't imported... Maybe I'm missing something

webapp/apps/dynamic/models.py
```diff
@@ -87,6 +87,27 @@ class DynamicBehaviorSaveInputs(models.Model):
...
+        else:
+            return rename_keys(self.tax_result, PRE_TC_0130_RES_MAP)
+
+        # older PB versions stored commit reference too
+        # e.g. taxcalc_vers = "0.9.0.d79abf"
+        if taxcalc_vers >= LooseVersion("0.13.0"):
```
codekansas: You could put this in a constant somewhere, like "taxcalc_min_compat_version" or something... Might be a bit clearer where this number comes from, so that people looking at it in the future are less confused.

webapp/apps/taxbrain/models.py
```diff
@@ -768,25 +769,26 @@ class TaxSaveInputs(models.Model):
...
         else:
-            return rename_keys(self._tax_result, PRE_TC_0130_RES_MAP)
+            return rename_keys(self.tax_result, PRE_TC_0130_RES_MAP)
 
-    @tax_result.setter
-    def tax_result(self, result):
-        self._tax_result = result
+        # older PB versions stored commit reference too
+        # e.g. taxcalc_vers = "0.9.0.d79abf"
+        if taxcalc_vers >= LooseVersion("0.13.0"):
```
codekansas: See comment above... Might also be a good idea to make a `utils.py` file for this commit anyway?

webapp/apps/dynamic/models.py
hdoupe: Nope, thanks for catching that. I added the import statement and a test for testing this method on the dynamic behavioral side.

webapp/apps/dynamic/models.py
hdoupe: Done, thanks.

webapp/apps/taxbrain/models.py
hdoupe: Ok, I moved this method to the bottom of `taxbrain/helpers.py`. The utility and helper functions need to be refactored and re-organzied. See issue #816

webapp/apps/taxbrain/views.py
```diff
@@ -953,7 +953,10 @@ def output_detail(request, pk):
...
             #Just need the error message from one failed job
             error_msgs = dropq_compute.dropq_get_results([failed_jobs[0]], job_failure=True)
-            error_msg = error_msgs[0]
+            if error_msgs:
+                error_msg = error_msgs[0]
+            else:
+                error_msg = "Error: stack trace for this error is unavailable"
             val_err_idx = error_msg.rfind("Error")
             error = ErrorMessageTaxCalculator()
             error_contents = error_msg[val_err_idx:].replace(" ","&nbsp;")
```
codekansas: This chunk of code seems pretty suspect... Ideally I guess there would be some easily-viewable logs, and you wouldn't need to store error messages in a database. Could to comment on the logic behind this? I.e. if the error messages aren't designed to be read by users easily, what's the purpose of storing them in the database / showing them to the user instead of logging them on the server side and presenting a more generic error message?

hdoupe: This error message is supposed to be a stack trace that is returned by the policy model (e.g. Tax-Calculator). When you hit previous runs that did not complete, you see the saved error message:
screen shot 2018-02-03 at 10 25 40 am

The stack trace is logged on the server side, too.

So, saving the error message in the database is more for the user than for us to introspect. The logs are a better source of info for that.

hdoupe:
@codekansas do you have any more suggestions before #814 is merged?

codekansas:
LGTM

Issue #815, opened 2018-01-29
Problems on the "What is TaxBrain?" page


martinholmer:
When users scroll to the bottom of the "What is TaxBrain?" page they see this:

![screen shot 2018-01-29 at 10 35 34 am](https://user-images.githubusercontent.com/12170745/35518691-3a6ed72e-04e0-11e8-94ff-5ec85bf3b519.png)

There are two problem here:

1. The asterisk does not "point" anywhere.  (Should there be an `*` character in front of the "These members have ..." section?)

2. Clicking on the " T.J. Alumbaugh, Continuum Analytics" link produces this page: [RESOLVED BY #818]

![screen shot 2018-01-29 at 10 25 37 am](https://user-images.githubusercontent.com/12170745/35518895-d29ac8f0-04e0-11e8-877b-2d6f6c59a692.png)

hdoupe:
@martinholmer Thanks for pointing this out.

@codekansas can you swap out T.J.'s link with a link to his github profile?
https://github.com/talumbau

@MattHJensen what is the asterisk supposed to refer to?

martinholmer:
Pull request #818 resolved the [second problem](https://github.com/ospc-org/ospc.org/issues/815#issue-292453914) reported in this issue, but not the first.  Are there any plans to either drop the `*` (which seems like the better option to me) or add a `*` at the start of the "These members ..." phrase?

MattHJensen:
I'm happy with dropping the `*`. (sorry I missed the question to me in January).

martinholmer:
Pull request #944 resolves issue #815

PR #818, opened 2018-02-01
Changed all instances of TJ Alumbaugh's email address

codekansas:
Fix for https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/815

Hmm... Turns out there's a lot of repeated things that are kind of a mess to maintain. I was thinking that a good way to unify the "text-heavy" parts of this project would be to use something like Jekyll, so that the texts of like the "about" parts could be written first in Markdown then built to HTML, which would be much cleaner than just writing everything in HTML (as it currently stands). For example, disclaimers and stuff that is used frequently throughout multiple apps could be written once, so that if they have to be updated somewhere down the line it is cleaner / safer than running some complicated grep. For example, you can put peoples' emails in a yaml file so that they can be updated / referenced more easily...

The Jekyll stuff wouldn't add very much technical complexity / likelihood of breaking things, since it's just a static markdown renderer (i.e. renders to HTML templates). Design would be something like:

Add a subdirectory in the project devoted to static includes (i.e. `jekyll-templates/`) with two subdirectories, `src/` and `generated/`. Put all the Jekyll text in the `src/` directory. Add the `generated/` directory as a Django template location. When Jekyll stuff gets changed, run the command

`jekyll build --source src/ --destination generated/`

Anyway... Just an idea. I didn't do it in this commit because refactoring all the current HTML to Markdown seemed like it would take a long time. But anyway here's the command I used for this:

```bash
grep -rl 'https://www.continuum.io/content/tj-alumbaugh' . | xargs sed -i "" 's/https:\/\/www\.continuum\.io\/content\/tj\-alumbaugh/https:\/\/github.com\/talumbau/g'
```

### DIFF
--- a/templates/dynamic/behavior.html
+++ b/templates/dynamic/behavior.html
@@ -100,7 +100,7 @@
             <p><strong>
             Core Maintainers (static modeling)*:
             <ul style = "list-style-type:none">
-            <li>- <a href = "https://www.continuum.io/content/tj-alumbaugh" target = "blank" >T.J. Alumbaugh, Continuum Analytics</a></li>
+            <li>- <a href = "https://github.com/talumbau" target = "blank" >T.J. Alumbaugh, Continuum Analytics</a></li>
             <li>- <a href = "http://www.nber.org/people/daniel_feenberg" target = "blank" >Dan Feenberg, National Bureau of Economic Research</a></li>
             <li>- <a href = "http://www.polsim.com/MRH_vita.pdf" target = "blank" >Martin Holmer, Policy Simulation Group</a></li>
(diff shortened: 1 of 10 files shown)


hdoupe:
@codekansas Thanks for taking care of this.

RE using Jekyll: I'm not familiar with it, but it seems like a neat tool. The html parts of the site (and the front-end in general) have needed some attention for awhile. Would you mind opening a separate issue re-stating why we should use jekyll and any other front-end changes that fit in with that?

Give me some time to look at this PR, and we'll just merge it within the current html framework.

hdoupe:
Thanks for the thorough update @codekansas. I see what you mean about having a bunch of duplicated text.

Issue #816, opened 2018-01-31
Create "core" app with utilities and other shared code


hdoupe:
There are several functions that are duplicated in the `dynamic` and `taxbrain` apps. Further, several variables, functions, and classes in `taxbrain` are imported into `dynamic`. I think we should move this code into a utilities app named something like "core" or "utils."

Once we've accomplished this first part, we should begin looking at code in the `views.py` and `models.py` modules and thinking about how we reduce the amount of redundant code.

cc @codekansas

hdoupe:
cc @kpinkelman

Issue #795, opened 2018-01-03
_EITC_MinEligAge is not an integer in TaxBrain GUI input


martinholmer:
In closed [Tax-Calculator issue 1809](https://github.com/open-source-economics/Tax-Calculator/issues/1809), TaxBrain user @siervicul reported this problem:

I'm using the TaxBrain interface, and when changing the minimum age from the default of 25 to 21, I get the error:
```
ERROR: _EITC_MinEligAge value 21.0 is not integer for 2017
```
Even when I change it back to 25 to try to clear the error message, the error becomes:
```
ERROR: _EITC_MinEligAge value 25.0 is not integer for 2017
```

hdoupe:
@martinholmer do you think #795 is related to open-source-economics/Tax-Calculator#1782?  Tax-Calculator may be doing a check to see if this is an integer and adding an error message before checking if it is a float.

martinholmer:
@hdoupe said:
> Tax-Calculator may be doing a check to see if this is an integer and adding an error message before checking if it is a float.

Tax-Calculator has never allowed a float value for an integer policy parameter since we added the `integer_value` and `boolean_value` fields to each parameter record in `current_law_policy.json`, which happened in [Tax-Calculator pull request 1585](https://github.com/open-source-economics/Tax-Calculator/pull/1585) that was in Tax-Calculator 0.12.0 released on October 20, 2017.

Here is the `_EITC_MinEligAge` record in `current_law_policy.json`:
```
    "_EITC_MinEligAge": {
        "long_name": "Minimum Age for Childless EITC Eligibility",
        "description": "For a childless filling unit, at least one individual's age ...",
        "section_1": "Refundable Credits",
        "section_2": "Earned Income Tax Credit",
        "irs_ref": "Form 1040, line 66a&b, step 4, instructions.",
        "notes": "",
        "row_var": "FLPDYR",
        "row_label": ["2013"],
        "start_year": 2013,
        "cpi_inflated": false,
        "col_var": "",
        "col_label": "",
        "boolean_value": false,
        "integer_value": true,                          <==========================
        "value": [25],
        "range": {"min": 0, "max": 125},
        "out_of_range_minmsg": "",
        "out_of_range_maxmsg": "",
        "out_of_range_action": "stop"
    },
```

hdoupe:
[...]

PR #822, opened 2018-02-06
Update to Django 1.9 and store parameters as JSON object

### LINKED ISSUE #811: Save model parameters as JSON objects

hdoupe:
This PR resolves #811 and implements the plan proposed in this [comment](https://github.com/OpenSourcePolicyCenter/PolicyBrain/issues/811#issuecomment-360830996).

Upgrading to Django 1.9 gives us access to the JSON column type which was included with Postgres 9.4 (the version used by PolicyBrain). Several important pieces of data that are either already JSON objects or are more usefully stored as JSON objects are:
1. raw input from the GUI
2. raw file input
3. parsed and cleaned GUI input
4. results from the model run

The goal of this PR is to store the raw input from the GUI as a JSON object. 

See Django release notes for [1.8](https://docs.djangoproject.com/en/2.0/releases/1.8/) and [1.9](https://docs.djangoproject.com/en/2.0/releases/1.8/) for further context on the changes requireed for the Django update.

This PR is based off of changes in #814.


hdoupe:
Commit [`80d715a`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/80d715add09925f54f2daf95488b665428bcb048) removes the `jsonfield` import. This library raises deprecation warnings for Django >= 1.9 and is incompatible with Django >= 1.0. 

Even when using the `jsonfields.fields.JSONField` class, if you have a postgres database of version less than 9.4, then the underlying postgres column is saved as type `text` instead of `json` or `jsonb`. See dmkoch/django-jsonfield#140 for more information.

That means that you have to do add a new migration that specifies that you want to use django.contrib.postgres.fields.JSONField in order to actually change the underlying postgres type to `json` or `jsonb`. I did this in commit [`d647e32 `](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/d647e320fa4d0e3f312b96737c911679c011a877)

This was tested with the following procedure:
1. Duplicate the current production version of the postgres database
2. Apply these migrations to the duplicated database
3. Make sure that the data did not change
Steps 1 and 2:
```
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ createdb -T pb_prod_fixed pb_prod_fixed_jsonfields_to_text_fix
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ echo $DATABASE_URL 
postgresql://localhost/pb_test_fixed_jsonfields_to_text_fix
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ export DATABASE_URL=postgresql://localhost/pb_prod_fixed_jsonfields_to_text_fix
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ python manage.py runserver
Performing system checks...

System check identified no issues (0 silenced).

You have unapplied migrations; your app may not work properly until they are applied.
Run 'python manage.py migrate' to apply them.

February 07, 2018 - 18:34:14
Django version 1.9, using settings 'webapp.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
^C(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ python manage.py migrate
Operations to perform:
  Apply all migrations: account, sessions, admin, flatblocks, register, dynamic, sites, auth, taxbrain, contenttypes, btax
Running migrations:
  Rendering model states... DONE
  Applying account.0004_auto_20170416_1821... OK
  Applying admin.0002_logentry_remove_auto_add... OK
  Applying auth.0007_alter_validators_add_error_messages... OK
  Applying btax.0003_auto_20180205_2151... OK
  Applying dynamic.0017_auto_20180205_2004... OK
  Applying dynamic.0018_auto_20180205_2025... OK
  Applying sites.0002_alter_domain_unique... OK
  Applying taxbrain.0062_auto_20180118_2120... OK
  Applying taxbrain.0063_auto_20180118_2125... OK
  Applying taxbrain.0064_auto_20180118_2133... OK
  Applying taxbrain.0065_auto_20180118_2137... OK
  Applying taxbrain.0066_taxsaveinputs_ald_alimonyreceived_hc... OK
  Applying taxbrain.0067_auto_20180119_1513... OK
  Applying taxbrain.0068_auto_20180119_1515... OK
  Applying taxbrain.0069_auto_20180205_2004... OK
  Applying taxbrain.0070_auto_20180205_2025... OK
```

Step 3: 
Script:
```
from sqlalchemy import create_engine
import pandas as pd
import numpy as np
import json
import traceback

def get_isnull_df(engine):
    isnull_query = """
    select url.id url_id, tsi.id id, url.unique_inputs_id unique_inputs_id,
    url.model_pk model_pk, tsi.tax_result is null as result_isnull, tsi.tax_result as tax_result,
    tsi.creation_date creation_date,
    url.taxcalc_vers taxcalc_vers, url.webapp_vers webapp_vers
    from taxbrain_taxsaveinputs tsi
    join taxbrain_outputurl url
    on tsi.id = url.unique_inputs_id
    where tsi.tax_result is not null
    and url.id %% 20 = 0
    order by url.id desc
    """
    print("running query...")
    isnull_df = pd.read_sql_query(isnull_query, con=engine)
    isnull_df.sort_values("unique_inputs_id", inplace=True)
    # isnull_df.to_csv('isnull_df.csv', index=False)
    print("returning...")
    return isnull_df

nonmigrated_engine = create_engine("postgresql://localhost/pb_prod_fixed")
migrated_engine = create_engine("postgresql://localhost/pb_prod_fixed_jsonfields_to_text_fix")

print("querying databases...")
nonmigrated_df = get_isnull_df(nonmigrated_engine)
migrated_df = get_isnull_df(migrated_engine)

print("quick assertion...")
nonmig_isnull, mig_isnull = nonmigrated_df.result_isnull.sum(), migrated_df.result_isnull.sum()
print("num not null for (nonmigrated, migrated): ",len(nonmigrated_df) - nonmig_isnull, len(migrated_df) - mig_isnull)
assert nonmig_isnull == mig_isnull

print("parsing json data...")
nonmigrated_df["topython_tax_result"] = nonmigrated_df.tax_result.apply(json.loads)
migrated_df["topython_tax_result"] = migrated_df.tax_result.apply(lambda x: json.loads(json.dumps(x)))
# print(nonmigrated_df[["tax_result", "topython_tax_result"]])
# print(migrated_df[["tax_result", "topython_tax_result"]])

print("starting assertion loop...")
for url_id, exp, act in zip(nonmigrated_df.url_id.values, nonmigrated_df.topython_tax_result.values, migrated_df.topython_tax_result.values):
    try:
        np.testing.assert_equal(act, exp)
        # print("passed: ", url_id)
    except AssertionError as ae:
        print("failed: ", url_id)
        traceback.print_exc()

```
Output:
```
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ python results_same_after_migration.py 
querying databases...
running query...
returning...
running query...
returning...
quick assertion...
('num not null for (nonmigrated, migrated): ', 1334, 1334)
parsing json data...
starting assertion loop...
(aei_dropq) HDoupe-MacBook-Pro:PolicyBrain henrydoupe$ 

```

I also checked the types in the database, `pb_prod_fixed_jsonfields_to_text_fix` to make sure that the type was changed from `text` to `jsonb`.

hdoupe:
So far, this PR has fully updated the TaxBrain app to Django 1.9 and mostly updated the rest of PolicyBrain. The only remaining part is updating the request payload data access to use the `GET` and `POST` methods.

Part 2 of this PR changes the way that user input is stored and accessed. The data will no longer be saved as individual fields. Instead, the data will be saved as JSON objects. One approach to doing this is:

1. Add a static method `add_fields` to the model form object, `TaxBrainForms` (formerly `PersonalExemptionForm`). This method filters out meta-data such as the creation date and stores the model parameters in a dictionary (commit [`1cfd34f`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/1cfd34fd7bdd21b0e42ab6db9be91aba7107600b))
2. Add a `set_fields` method to the model, `TaxSaveInputs`. This method filters out the parameters that were chosen by the user, maps the parameter names from TaxBrain style to Tax-Calculator style, and converts the user input into the type specified by Tax-Calculator for that particular parameter (partially resolving #786 and resolving #795) (stub added in commit [`8ac6520`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/8ac6520ba359d21fa9ad0d0c61437b0f1464b9c4))
3. Add a `get_model_specs` method to the model, `TaxSaveInputs`. This method converts the cleaned data in the `fields` attribute (built in `set_fields`) to the format expected by the Tax-Calculator model-running end-points (stub added in commit [`8ac6520`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/8ac6520ba359d21fa9ad0d0c61437b0f1464b9c4))

Note that (2) and (3) are largely a refactoring of the logic that is kicked off when [`get_reform_from_gui`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/webapp/apps/taxbrain/views.py#L247-L315) is called from [views.py](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/webapp/apps/taxbrain/views.py#L449-L451).

I'm interested in others' thoughts on this approach. 

cc @talumbau

hdoupe:
@kPinkelman commit [`eff36a5`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/eff36a5c826f32898ae4ec47626730e5d069b93c) is relevant to your work on #816

hdoupe:
Status update: 

- [x] Update to Django 1.9
- [x] Save all inputs as JSON object for taxbrain-static and taxbrain-partial equilibrium apps
- [ ] Preserve backwards compatibility for loading parameters from previous runs. Tax-Calculator 0.15.0 renames several characters. For example, `PT_exclusion_wage_limit` was renamed to `PT_excl_wagelim_rt`. I'll give more thought to how we can handle this in a comment below. Note: this will not affect the user's ability to view previous results. See #811 for more information on this distinction.
- [ ] Create policy for maintaining backwards compatibility in loading parameters of old (something like 4-6 months old) runs. This will probably be discussed in an on-going basis. 
- [ ] Upgrade to a B-Tax version that is compatible with Tax-Calculator 0.15+. I believe that this [version](https://github.com/open-source-economics/B-Tax/releases/tag/0.2.1) has been released but is not yet on Anaconda.

As soon as soon as the next version of B-Tax is released, I will push this PR to the test server where it needs to be tested very thoroughly. During this time, I will work on providing backwards compatibility for loading parameters from previous runs.

hdoupe:
@martinholmer Commit [260dc45](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/260dc45720d2cf3a31fa562df57e37d84517385f) adds logic that converts raw input strings into the Python type that Tax-Calculator expects. You'll notice that it "fails silently" and relies on Tax-Calculator to provide error messages. If, for example, a `float` like 22.2 is received but TC expects type `int` then PolicyBrain does not cast it down to 22 because that is a sneaky, unexpected behavior that the user would find suprising in my opinion. I'm interested in your thoughts both on how I could clean it up (it turned into a gnarly mass of if statements) and on how it performs.

The relevant function is [`parse_value`](https://github.com/hdoupe/PolicyBrain/blob/update_django19/webapp/apps/taxbrain/param_formatters.py#L84-L179) and the tests for it are [here](https://github.com/hdoupe/PolicyBrain/blob/update_django19/webapp/apps/test_assets/test_param_formatters.py#L85-L125).

martinholmer:
@hdoupe said in a pull request #822 status update:
> Preserve backwards compatibility for loading parameters from previous runs. Tax-Calculator 0.15.0 renames several characters. For example, `PT_exclusion_wage_limit` was renamed to `PT_excl_wagelim_rt`. I'll give more thought to how we can handle this in a comment below.

You should check with @codykallen about this, but my understanding is that `PT_excl*` variables that were renamed were renamed because the logic associated with the old names was not quite correct.  So, I'm not sure you can or want to "preserve backwards compatibility for the old `PT_exclusion*` variables.

codykallen:
@martinholmer said:

> my understanding is that PT_excl* variables that were renamed were renamed because the logic associated with the old names was not quite correct. So, I'm not sure you can or want to "preserve backwards compatibility" for the old PT_exclusion* variables.

That is correct. The original implementation of the wage limitation on the pass-through income exclusion was incorrect. The original implementation assumed that it would automatically apply if someone used the pass-through income exclusion, but the law actually phases in the wage limitation. I changed the name of the parameter (1) because the logic for the wage limitation changed and (2) because the parameter names for that were becoming very long and shortening 'exclusion' to 'excl' reduced problems with overly long lines of code. 

Given that the original implementation of the pass-through exclusion was incorrect (and the implementation has never been proposed), I don't think it's worthwhile to retain the 'PT_exclusion*` variables.

hdoupe:
@martinholmer @codykallen Thanks for your input. I think the qualitative concerns you outlined warrant some more thought. 

When trying to display user input on the GUI page from older runs, do you think that reasonable behavior would be to try to look up the old parameter names, and, if they are not there, display warning messages at the top of the screen saying something like: 

"Parameters param_x, param_y have been deprecated. Refer to the Tax-Calculator Documentation for sensible replacements."

cc @MattHJensen

martinholmer:
@hdoupe said:
> When trying to display user input on the GUI page from older runs, do you think that reasonable behavior would be to try to look up the old parameter names, and, if they are not there, display warning messages at the top of the screen saying something like:
>
> "Parameters param_x, param_y have been deprecated. Refer to the Tax-Calculator Documentation for sensible replacements."

This sounds to me like a very sensible approach.  I'm not sure what the other options would be.

hdoupe:
@martinholmer Thanks for the code review. The minor fixes sound good to me. The most recent commits incorporated them. 

@martinholmer  said
> Is there a Python option other than `ast.literal_eval()`?
If so, you could eliminate all the special handling of negative numbers, which complicates the code.

It turns out that `ast.literal_eval()` works for negative numbers. I'm not sure why I was having an error when I tried to use that for negative numbers last week.

> A quick review suggests you are trying to help TaxBrain users by correcting their mistakes (like entering FAlse instead of False), right? That's an admirable goal, but when pursuing that goal on past projects I found that the "helping" code gets quite complex. There are lots of ways users can make mistakes and anticipating (and correcting) them is inherently complex.

I refactored the approach to casting in commit ([`cb0f556`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/cb0f556fadf79130bc65e36f28720facdad96e33). Hopefully, that makes the code a little more clear. 

I agree that the helping code can become pretty complex and that we should try to do less and simply display the errors to the user. There are some cases that may be confusing to the non-technical user. For example, why does `True` work when `true` does not? I think PolicyBrain should correct that. But, entering 22.0 instead of 22 for "EITC Minimum Eligibility Age" is an interesting case. I'm not sure if we should cast that to 22 or not. Regardless, for some reason, PolicyBrain displays the default value for this field as `22.0` and I don't know why. Until we fix that, we probably need to cast it into an `int`. 

I will continue to tinker with the logic in this function. Hopefully, we can remove the final set of if/else statements and just return the value from `ast.literal_eval`.

hdoupe:
PR #822  is on the test app at http://ospc-taxes7.herokuapp.com/.

I encourage everyone to do their best to break this changes pushed by this PR. This is a really big release and a large portion of the TaxBrain internals were either changed or refactored. 

[Run 1533](http://ospc-taxes7.herokuapp.com/taxbrain/edit/1533/?start_year=2018) is a good example for checking out the new parameter deprecation error. Note that only parameters that have changed since the previous version of PolicyBrain qualify for this message. The policy up to now has been to rename columns to what we thought was the parameter's new name.

Note that CCC is down on the test app until we figure out how to update the anaconda package.

Also, I am now having trouble loading runs submitted before 2016, but the problems described in #812 and #813 are resolved.

I will update the RELEASES.md file with a comprehensive list of the changes going into this release.

hdoupe:
@MattHJensen @martinholmer commit [`f95a420`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pull/822/commits/f95a4207518a22c1e0f184cea079e04b8f8b9463) adds a document describing PolicyBrain's approach to parsing parameters and displaying deprecated parameters. I'm interested in any thoughts on this that you may have.

martinholmer:
@hdoupe said:
> commit f95a420 adds a document describing PolicyBrain's approach to parsing parameters and displaying deprecated parameters. I'm interested in any thoughts on this that you may have.

Look reasonable to me. 
Thanks for the massive amount of work you've done on this highly sensible improvement!

hdoupe:
@martinholmer and @MattHJensen Thanks for the review. I plan to merge this and do a release as soon as I resolve some minor issues with CCC.

PR #832, opened 2018-03-06
Containerize worker nodes

hdoupe:
This PR uses Docker to containerize the worker nodes. This gives the following advantages:

1. Only have to build the updated environment **once**! This means that you don't have to build the environment locally for local developement and **ten** more times for all of the worker nodes.
2. Local development environment is **very close to identical** to the production environment.
3. We can start to phase in container orchestration software which will make deployments **faster** and **safer** and allows us to **scale** the number of worker nodes as needed.

To-do list for remaining items coming shortly...


hdoupe:
Docker setup documentation will be addressed next week in a follow up PR.

PR #849, opened 2018-03-16
Add CCC start year drop-down

hdoupe:
This PR adds a drop down menu for the start year in the CCC. 

@GoFroggyRun I'm having trouble getting the java script code to link up with the html code. That is the dialogue box doesn't change when I change the start year. Do you have any suggestions?

Also, I removed the staticfiles and fixed the gitignore so that they aren't stored anymore.

cc @jdebacker


static/js/btax.js
```diff
@@ -31,7 +31,13 @@ $.each(asset_yr_str, function(index, value){
   });
 });
 
-$('#start-year-select').change(function() {
-  $('#current-year-link').attr('href', '/ccc/?start_year=' + $(this).val());
-  $('#current-year-alert').removeClass('hidden');
+var currentYear = $('#start-year-select').val();
+$('#start-year-select').change(function(e) {
+  $('#current-year-link').attr('href', '/taxbrain/?start_year=' + $(this).val());
```
GoFroggyRun: @hdoupe After changing `'/taxbrain/?start_year='` to `'/ccc/?start_year='` on my local, I was able to have the dialogue popping up and working. 

screen shot 2018-03-16 at 1 05 59 pm

hdoupe: Nice, thanks @GoFroggyRun. Everything is looking good now.

PR #853, opened 2018-03-20
[WIP] Add age button to "build table" widget 

GoFroggyRun:
This PR adds an extra button to "build table" widget. The button set works similar to our current "payroll vs income" button, in the way that users could select either of them or both. When both "Under Age 65" and "Age 65 & Above" are selected, the table will render full data; when either one is selected, the respective data source would be rendered. Here's a preview of the button:

screen shot 2018-03-20 at 2 28 06 pm

This PR needs further testing after desired data source is incorporated.


hdoupe:
Looks good @GoFroggyRun. Thanks for putting this together. This is on hold until the necessary Tax-Calculator changes are made to provide tables with an age breakdown.

PR #856, opened 2018-03-21
Add link to data document

andersonfrailey:
Add a link to the data description doc to TaxBrain. Full text is:
>TaxBrain now offers users two data options: CPS and PUF. Both datasets contain tax unit level data, but they do not have all of the same variables. As a result, you will not be able to change some policy parameters, depending on which dataset you choose. A more detailed breakdown of the differences between the file can be found here.

cc @hdoupe @MattHJensen


hdoupe:
LGTM. Here's the what the top of the TaxBrain inputs page looks like now:

screen shot 2018-03-21 at 1 49 45 pm

PR #860, opened 2018-03-22
Restore "Print" button for elasticity output page

GoFroggyRun:
I removed all non-functional buttons for the elasticity result page for the moment. I also plan to permanently get rid of the "Save As" drop-down whose options don't make much sense in my opinion.

I re-organized the div element a bit so that the table can be properly accessed, and was able to have the "Print" button semi-working. Here's how the page would look like:

screen shot 2018-03-22 at 4 36 24 pm
A new window would pop-out with the same content in the table that allows you to print/save. I plan to polish the table layout a bit which would complete this button.

### DIFF
--- a/templates/dynamic/elasticity_results.html
+++ b/templates/dynamic/elasticity_results.html
@@ -126,31 +126,24 @@
       <a href="/dynamic/macro/edit/{{ unique_url.pk }}/?start_year={{ first_year }}" class="text-white btn btn-secondary">Edit Parameters</a>
     </div>
 
-    <div class="result-table">
       <div class="result-table-controls">
         <div class="controls-tools">
-          <div class="btn-group">
-            <button type="button" class="btn btn-text btn-sm dropdown-toggle" data-toggle="dropdown" aria-expanded="false">
-              Save As <span class="caret"></span>
-            </button>
-            <ul class="dropdown-menu" role="menu">
-              <li><a href="/taxbrain/pdf">PDF Report</a></li>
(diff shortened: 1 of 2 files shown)


martinholmer:
@GoFroggyRun proposed this Macro Elasticity Simulation output table:

screen shot 2018-03-22 at 4 36 24 pm

Give the row label on the left, I wound seem as if the row of `%` symbols below the top year row is redundant.

GoFroggyRun:
@martinholmer said:
>Give the row label on the left, I wound seem as if the row of % symbols below the top year row is redundant.

Thanks for pointing this out @martinholmer. I agree that the extra row of % seems redundant. Would it be sensible to eliminate the row in the original table as well? I.e., modifying the table to something like:

screen shot 2018-03-22 at 5 15 16 pm

Personally, I find the new table more concise yet as informative as the old one.
cc @hdoupe @MattHJensen

martinholmer:
@GoFroggyRun,  Your latest version of the table looks good.  Thanks.

hdoupe:
Thanks @GoFroggyRun. The new version of the table looks good to me, too.

hdoupe:
@GoFroggyRun Can you push the changes that remove the percent symbol? Should be good to go once you do that.

GoFroggyRun:
@hdoupe thanks for the reminder. I was hoping to include "Copy" button as well in this PR. However, its layout seems a little bit tricky. I think it's a good idea to merge this PR first. I'll have "copy" button in another PR once finished.

hdoupe:
LGTM. Thanks @GoFroggyRun for the contribution and @martinholmer for the review!

PR #870, opened 2018-03-27
Update RELEASES.md and increase compute time to 100 seconds

hdoupe:


hdoupe:
I still had some issues with the compute time at 90 seconds. Most of the runs finished in under 90 seconds, but some are close to 100 seconds. Thus, we still had some of the issues with the loading page hanging. Hopefully, this gets the ETA right.

PR #872, opened 2018-03-27
Change Dropdown Order

andersonfrailey:
This PR changes the order of the Data Source dropdown in TaxBrain so that `PUF` is the first option.


hdoupe:
LGTM. Thanks @andersonfrailey!

PR #874, opened 2018-03-27
[WIP] Add special HTML pages for when a model result is not found or is incompatible

### LINKED ISSUE #869: Render a special HTML page for when a model result is not found or is incompatible
For cases where the result is not found or is incompatible with existing code, I think we should display a page with an explanation for what has happened and how the result could potentially be either reproduced or re-run on the newer model. The message would be something like:

"We are unable to display the results for model run #XXX. We apologize for the inconvenience and are working to provide a satisfactory solution for displaying and/or recovering results created using older versions of PolicyBrain. Here is one option for recreating your results:

Go to the following link: http://www.ospc.org/taxbrain/edit/XXX/?start_year={{fill_in_start_year}} and re-submit your parameters using the current version of PolicyBrain. Note that for older runs some parameters may have been deprecated. We recommend the upstream package documentation for alternative parameter choices.

The upstream package may have more options for re-creating results with the exact version of the package used in this run."

cc @MattHJensen @GoFroggyRun

GoFroggyRun:
This PR tries to close #869. Two HTML pages are added to TB and CCC respectively, where the content of the page suggest an option to re-create outdated results.


GoFroggyRun:
Here's a preview for TB version: 
screen shot 2018-03-27 at 5 41 15 pm

@hdoupe could you take a look at these pages? Feel free to modify the content if you'd like. 

Would these be sufficient for you to render in `view.py`?

GoFroggyRun:
I also noted that CCC and TB are sharing the same header: see [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/templates/btax/not_ready.html#L6). Does it make sense to break TB's [header](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/templates/taxbrain/header.html) into two pieces, one for TB and the other for CCC, and place them in their respective directories?

hdoupe:
@GoFroggyRun these look good. I think we can go ahead and merge this. I'll update the back-end so that these pages are rendered on exceptions. I'll also tweak the language some. Thanks for the quick response on this PR.

I think it does make sense to break up the headers. Would you mind opening a separate PR to do this? There's a lot of apps referencing code in other apps in both in the front-end and in the back-end. I think it's a good idea to draw a clear line between the different apps and pull the shared code into a central place.

PR #881, opened 2018-04-02
Create respective `header.html` files for btax and taxbrain

GoFroggyRun:
Per discussion in #874, this PR separates the header template so that btax and taxbrain would use their own `header.html.` This makes future maintenance easier when content in either file needs to be updated. 

I noticed that there are some other btax HTMLs using files in taxbrain's directory. For instances [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/2fa70f4dd6f1d003208b4345e99bcc41befeff7a/templates/btax/includes/params/inputs/macro_econ.html#L1) and [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/2fa70f4dd6f1d003208b4345e99bcc41befeff7a/templates/btax/includes/params/inputs/macro_econ.html#L26-L27). In fact `taxbrain /input_form_section.html` and `taxbrain/includes/params/inputs/param.html` are extensively used for files within the directory `templates/btax/includes/params/inputs/`. I didn't make separate copies of the two as I think it seems a bit redundant and might make future maintenance even harder when `param.html` or `input_form_section.html` needs to be updated. Maybe we can rearrange the file/directory structure a bit to resolve this problem. 

@hdoupe does it make sense? Could you review this PR?

### DIFF
--- a/templates/btax/header.html
+++ b/templates/btax/header.html
@@ -1,0 +1,79 @@
+{% load staticfiles %}
+{% load flatblocks %}
+<div id="affix-offset">
+  <nav class="logobar" role="navigation">
+    <a class="logobar-brand" href="/">
+      <img alt="TaxBrain: A disruptive open source platform for exploring tax policy impact" src="{% static 'images/logo-no-icon.png' %}">
+    </a>
+  </nav>
+  <section class="taxbrain-hero taxbrain-hero-small">
+    <div class="taxbrain-hero-block">
+        <div class="logo">
+          <h2 class="sr-only">TaxBrain</h2>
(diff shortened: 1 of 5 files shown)


hdoupe:
@GoFroggyRun Thanks for the PR. This looks good to me.

I think re-arranging the `templates` directory is a good idea. Do you want to start that in a separate PR?

GoFroggyRun:
@hdoupe thanks for reviewing. Yes, I plan to work on re-arrangements in another PR.

hdoupe:
No problem. Ok, sounds good.

PR #886, opened 2018-04-24
Convert to Python 3

hdoupe:
This PR converts PolicyBrain to Python 3. So far, I have only run the Python 2to3 module. I'm pretty impressed with the initial results...only 22 out of the 172 tests failed. However, even once all the tests are passing, we will have to take care to test the app thoroughly before deploying it.


hdoupe:
The heroku buildpack needs to be updated, too. I plan to do this in a follow up PR. We currently use a [build-pack](https://github.com/OpenSourcePolicyCenter/conda-buildpack/) that was forked from the official buildpack: https://github.com/heroku-python/conda-buildpack.

Also, the worker nodes need to be upgraded to Python 3. Note that the webapp and the worker nodes can be out-of-sync since the worker nodes run as separate processes. I will update this in a follow up PR.

PR #888, opened 2018-04-30
Remove no longer used gevent script

hdoupe:
As far as I can tell, the `custom_work.py` script is not used. I'm not sure how it was used previously either which makes me a little nervous about simply removing it. @talumbau is it safe to remove this script?


hdoupe:
Ok, thanks for letting me know.

PR #895, opened 2018-06-12
Rename CNN.jpg to cnn.jpg so CSS can find it

hdoupe:


hdoupe:
I thought this was fixed in #889. It turns out git's `diff` functionality is case-insensitive when it comes to file names. So, this change was never displayed in the `git diff` and I did not double check to make sure the change was checked in. You can force git to be case-sensitive with regard to filenames via the `git mv` command, e.g. `git mv CNN.jpg cnn.jpg`. Anyways this is resolved now.

PR #902, opened 2018-06-14
Add necessary steps to setup instructions

lucassz:
I noticed while setting up a local installation of PolicyBrain that these steps are definitely needed (to install prerequisite packages TaxBrain and B-Tax).

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Local Deployment Setup
 
 First, if you plan on contributing to PolicyBrain, then fork PolicyBrain and work off of that fork.  If you do not plan to contribute, then you can clone the main repo.
 
-Make sure that [Anaconda/Miniconda](https://www.anaconda.com) are installed. Make sure that the local Postgres Server is running. Then, open a terminal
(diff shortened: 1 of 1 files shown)


hdoupe:
@lucassz Thanks for opening up this PR. The instructions can always be improved. 

I have a question about this line:
> conda install -c ospc taxcalc btax

These packages should be installed in the `python_env_build.sh` script. Were they not in the `pb_env` conda environment?

lucassz:
> These packages should be installed in the `python_env_build.sh` script. Were they not in the `pb_env` conda environment?

@hdoupe I remember having to perform that step manually even in the environment, but you are right that they should have been installed according to the `condo-requirements.txt` file.

hdoupe:
What's the output when you run `source activate pb_env && conda list taxcalc`?

lucassz:
It seems to show up fine in the case, so it's possible I was doing something wrong before.

hdoupe:
Ok, cool. In that case, you can remove the extra `conda install` step and we can merge this PR with the link addition.

lucassz:
@hdoupe Done.

hdoupe:
LGTM. Thanks for the contribution @lucassz

PR #903, opened 2018-06-14
Various code style improvements, mainly to TaxBrain views

lucassz:
Various suggestions for changes to the TaxBrain views file, mainly to meet PEP8 guidelines and for general cleanup. Also, some list comprehensions are changed to generator expressions. I'm looking forward to feedback about whether these are too many or too few changes for this purpose.

### DIFF
--- a/webapp/apps/taxbrain/views.py
+++ b/webapp/apps/taxbrain/views.py
@@ -271,10 +271,10 @@ def submit_reform(request, user=None, json_reform_id=None):
 
     # We need to stop on both! File uploads should stop if there are 'behavior'
     # or 'policy' errors
-    warn_msgs = any((len(errors_warnings[project]['warnings']) > 0
-                     for project in ['policy', 'behavior']))
-    error_msgs = any((len(errors_warnings[project]['errors']) > 0
-                      for project in ['policy', 'behavior']))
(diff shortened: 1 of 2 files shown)


webapp/apps/taxbrain/views.py
```diff
@@ -801,26 +802,28 @@ def output_detail(request, pk):
...
             model.error_text = error
             model.save()
-            return render(request, 'taxbrain/failed.html', {"error_msg": error_contents})
-
+            return render(request, 'taxbrain/failed.html',
+                          {"error_msg": error_contents})
 
-        if all([j == 'YES' for j in jobs_ready]):
+        if all((j == 'YES' for j in jobs_ready)):
             results = dropq_compute.dropq_get_results(normalize(job_ids))
```
hdoupe: @lucassz what's your motivation for using a tuple instead of a list?

hdoupe: Ah, neat. Yep, looks like you don't need the second-set of parentheses.

hdoupe: Thanks for the explanation.

hdoupe:
LGTM, thanks for the contribution @lucassz

Issue #885, opened 2018-04-23
Downloaded CSV files are not rectangular


martinholmer:
When importing a downloaded CSV file into a spreadsheet, it is so distorted it is useless.

The distortion is caused by fact that the first (title) line in the downloaded CSV file is not comma separated.  This is strange because the last (source) line is comma separated, as is the body of the table.

What's the schedule for fixing this serious bug?

@hdoupe @GoFroggyRun

hdoupe:
@martinholmer Thanks for the bug report. I was able to open the CSV in Excel and Numbers (the Apple version of Excel). That being said, this is a CSV and not an Excel spreadsheet. Thus, the odd formatting is not helpful. I'm taking stock of the open issues today and will open an issue to outline PolicyBrain priorities for the near future.

martinholmer:
@hdoupe said:
>  I was able to open the CSV in Excel and Numbers (the Apple version of Excel). That being said, this is a CSV and not an Excel spreadsheet. Thus, the odd formatting is not helpful. 

Yes, I too can open up the downloaded CSV files in my spreadsheet, but the non-rectangular format leaves the spreadsheet misaligned.

It's not just spreadsheets that will be processing downloaded CSV files.  And those other tools will be expecting a rectangular file.  That is, one that has the same number of comma-separated values per row.
The current TaxBrain CSV files are not like that.

hdoupe:
Thanks @martinholmer. I agree. 

I haven't done anything with the CSV file code before. I'm having trouble figuring out where they are even created. @GoFroggyRun are the CSV files built with javascript?

GoFroggyRun:
@hdoupe yes. These buttons are within the [`datatables.js`](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/static/js/vendor/DataTables/datatables.js) file. And the function related to CSV output button can be found [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/static/js/vendor/DataTables/datatables.js#L27679).

hdoupe:
@martinholmer #885 should have been fixed in PR #904. Are you interested in testing out this fix? You can verify it at https://ospc-test.herokuapp.com/

martinholmer:
@hdoupe said:
> #885 should have been fixed in PR #904. Are you interested in testing out this fix? You can verify it at https://ospc-test.herokuapp.com/

Sorry, I don't have time.

[...]

PR #904, opened 2018-06-14
Make title row of generated CSV file have correct number of columns

lucassz:
Addresses #885. It may be that the minified file has to be updated too, although I'm not sure if it's actually being used.

### DIFF
--- a/static/js/vendor/DataTables/datatables.js
+++ b/static/js/vendor/DataTables/datatables.js
@@ -27693,13 +27693,14 @@ DataTable.ext.buttons.csvHtml5 = {
 		var output_ = _exportData( dt, config ).str;
         var data = dt.buttons.exportData( config.exportOptions );
         if (data.header[2][0] == 2) {
-            var str = $('tr:first').text() + '\n'
-            var output = str + output_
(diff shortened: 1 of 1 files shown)


static/js/vendor/DataTables/datatables.js
```diff
@@ -27693,13 +27693,14 @@ DataTable.ext.buttons.csvHtml5 = {
...
-            var output = str + output_
+            var str = $('tr:first').text();
+        }
+        else {
+            var str = $('h1:last').text();
         }
+        var commas = (output_.split('\n')[0].match(/,/g) || []).join('');
+        var output = '"' + str + '"' + commas + '\n' + output_;
+
         var url = window.location.href;
```
hdoupe: @lucassz I'm having trouble figuring out what's going on here. Would you mind adding a comment or two?

lucassz: Added now, hope that's clearer. The first line is substantially different from the last line because the first line is a caption, whereas the last line actually is part of the body of the table. This manually makes it as though the caption were another row.

PR #907, opened 2018-06-18
Implement parameter validation for input_fields

lucassz:
Implement parameter validation for input_fields, both for TaxBrain and for the partial equilibrium model.

### DIFF
--- a/webapp/apps/taxbrain/tests/test_views.py
+++ b/webapp/apps/taxbrain/tests/test_views.py
@@ -203,9 +203,6 @@ class TestTaxBrainViews(object):
                             str(START_YEAR), data_source=data_source)
 
     def test_taxbrain_post_no_behavior_entries(self):
-        # marking as xfail for 822--don't have a way to
-        # check if too invalid parameters are posted
-
         #Monkey patch to mock out running of compute jobs
         get_dropq_compute_from_module('webapp.apps.taxbrain.views')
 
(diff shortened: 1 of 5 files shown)


hdoupe:
@lucassz This looks good. Can you add a test demonstrating the validation behavior? There aren't any specific tests for the forms. You can create a new file `taxbrain/tests/test_forms.py` with the new tests. A test with OK parameters and a test with unknown parameters should be sufficient. 

You can look to `webapp/apps/test_assets/test_models.py` and `webapp/apps/taxbrain/tests/test_models.py` for some relevant examples.

lucassz:
@hdoupe Done. It seemed more relevant to add it to `test_views.py` since there was similar code there already.

webapp/apps/taxbrain/tests/test_views.py
```diff
@@ -57,6 +57,20 @@ def test_taxbrain_post(self, data_source):
         check_posted_params(result['tb_dropq_compute'], truth_mods,
                             str(START_YEAR), data_source=data_source)
 
+    @pytest.mark.xfail
```
hdoupe: @lucassz This shouldn't be xfail-ed. That means that `py.test` will ignore it if the test fails.

lucassz: @hdoupe Thank you. I think I had understood the purpose of `xfail`.

hdoupe:
@lucassz This looks great. Thanks for taking care of this and for finding a much simpler way to solve the problem.

hdoupe:
@lucassz This should be good to go once you knock out the merge conflicts.

lucassz:
@hdoupe Happy to do it. Should be good to go now.

hdoupe:
Great, thanks @lucassz

PR #905, opened 2018-06-14
Allow for running the model without reforms

lucassz:
Addressing second part of #883, i.e. removing warnings/errors for running the TaxBrain model with current tax policy / no reforms.

Issue #883, opened 2018-04-03
Regressions in TaxBrain GUI interface


martinholmer:
With the new 1.5.0 version of PolicyBrain, the TaxBrain GUI input page has suffered two significant regressions.

1. Users cannot scroll down the left-hand pane to get quickly to sections or the "Show Me the Results" button. [UNRESOLVED AS OF 2018-Nov-06]

2. Users cannot simulate current-law policy because TaxBrain now forces users to enter a reform. [RESOLVED BY #905]

martinholmer:
Pull request #905 resolved the [second problem](https://github.com/ospc-org/ospc.org/issues/883#issue-310907728) originally reported in this issue, but the first problem remains unresolved.

hdoupe:
Thanks for re-upping the second problem @martinholmer.

martinholmer:
Presumably TaxBrain issue #883 will be fixed when switching to the new webapp infrastructure.

PR #908, opened 2018-06-18
Add default parameter handling for macro elasticity simulation

lucassz:
Add default parameter handling for macro elasticity simulation. Follow-up to #883 and #905: currently, running the macro elasticity model without inputting any data results in not even a UI error, but an uncaught server error in converting an empty string to a float. This adds default handling in a general way. However, it does make the assumption that there is only one column/value per input; this was already implied in the previous code, and I believe that should be okay for this model, but please let me know if it is not.

### DIFF
--- a/webapp/apps/dynamic/tests/test_views.py
+++ b/webapp/apps/dynamic/tests/test_views.py
@@ -1,6 +1,5 @@
-
-from django.test import TestCase
 from django.test import Client
+import pytest
 import mock
 import os
 # os.environ["NUM_BUDGET_YEARS"] = '2'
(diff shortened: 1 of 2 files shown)


hdoupe:
@lucassz Nice catch. Would you mind adding a test in the `dynamic`app where a reform is submitted with the defaults?

lucassz:
@hdoupe This is done. I had to make the class not be a TestCase to get the parametrization to work.

webapp/apps/dynamic/tests/test_views.py
```diff
@@ -15,13 +14,19 @@
...
     def setUp(self):
         # Every test needs a client.
         self.client = Client()
 
+    def assertEqual(self, a, b):
+        assert (a == b)
+
+    def assertTrue(self, a):
+        assert a
+
```
hdoupe: @lucassz what do you think about dropping the `assert*` methods? When I removed the `TestCase` inheritance from some of the other test classes, I dropped these methods because I think it's easier to just read `a==b` instead of `self.assertEqual(a, b)`.

lucassz: Yes, I think that is a better idea. I was trying to minimize changes to the code, but will make the change.

hdoupe:
LGTM. @lucassz are you ready for this to be merged?

lucassz:
@hdoupe Seems good to me too.

hdoupe:
Cool, thanks for the contribution.

PR #909, opened 2018-06-18
Change datetime objects to timezone-aware ones

lucassz:
Change naive datetime objects to timezone-aware ones. This has the advantage of fixing 200+ warnings when running the tests, and everything seems to work with this change.

### DIFF
--- a/webapp/apps/btax/models.py
+++ b/webapp/apps/btax/models.py
@@ -10,6 +10,7 @@ from django.contrib.auth.models import User
 
 from django.contrib.postgres.fields import JSONField
 import datetime
+from django.utils.timezone import make_aware
 
 from ..taxbrain.models import (SeparatedValuesField,
                                CommaSeparatedField)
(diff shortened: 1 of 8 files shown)


hdoupe:
@lucassz This is extremely helpful. Thanks for knocking this out. Just so I understand, `timezone.now()` is the correct way to get the date when the time zone support is enabled and `make_aware` is a helper function for compatibility?

lucassz:
@hdoupe Right, yes. We just need to use `make_aware` in order to adapt the hardcoded times that are used as defaults (not sure whether those are absolutely needed either).

hdoupe:
Ok, so right now, `TZ` is set to `UTC`. It seems like the old times were set to some other timezone and `time_aware` converts them to `UTC`. 

I'm reading the django timezone docs now. I'm having trouble figuring out which timezone we were using before when we had `USE_TZ` set to `True`, `TIME_ZONE` set to `UTC`, but were not using timezone aware datetimes. Do you have any ideas on this?

lucassz:
@hdoupe According to my understanding of the Django documentation, it would indeed default to the `TIME_ZONE` in `settings.py`, in this case `UTC`. So the behavior is not changing -- wrapping it in `make_aware` is just doing explicitly what Django was doing implicitly (with a warning) in a way that's deprecated but supported to preserve backward compatibility.

hdoupe:
Ok, that makes sense. Let's just leve the `make_aware` wrapper to be on the safe side. Is this ready to be merged?

lucassz:
@hdoupe Yes. Plus it gets rid of the warnings since we're doing it explicitly, so the Travis output will be cleaner.

hdoupe:
Very nice. Thanks for the contribution.

PR #910, opened 2018-06-19
Worker node cleanup

hdoupe:
This PR performs some maintenance work in the flask app for the worker nodes:
- [x] Use celery to check job status not a list that is kept in memory
- [x] Refactor flask app structure to follow this [popular example](https://github.com/pallets/flask/tree/1.0.2/examples/tutorial)
- [x] Convert to Python 3
- [x] Write tests to cover flask endpoints
- [x] Use msgpack for project inputs serialization [EDIT]


hdoupe:
@lucassz would you mind reviewing PR #910?

distributed/taxbrain_server/_version.py
```diff
@@ -39,19 +39,19 @@ def run_command(commands, args, cwd=None, verbose=False, hide_stderr=False):
             if e.errno == errno.ENOENT:
                 continue
             if verbose:
-                print("unable to run %s" % args[0])
+                print(("unable to run %s" % args[0]))
```
lucassz: As we're moving to Python 3, maybe it would be better to use the new `format(…)` syntax which is now preferred over `%`?

hdoupe: Right, I wound up deleting this file. I don't think there are any other places where strings are formatted.

hdoupe:
Next steps:
- Develop testing harness such that docker is not necessary for running tests. The best way to run the tests no is to run: `docker-compose run flask py.test -s -v`. I've found that re-building docker images each time tests are run is a little burdensome. Plus, running celery in a different process means that you can't mock out the celery calls.
- Refactor celery functions to have a standard data flow. Currently, there is still some ad hoc code for each celery task. I think that it can be standardized a little more.

Issue #913, opened 2018-06-25
Automate deployment and simplify infrastructure


lucassz:
Per a discussion with @hdoupe, I plan to work on the automation of the deployment that's currently done manually for both the webapp and backend worker nodes. Here are the changes on which I am planning, in order of least to most different from the current setup:

1. Automate the deployment of Heroku and AWS instances through [Terraform](https://www.terraform.io), a tool for declaratively describing infrastructure through code and orchestrating its deployment. This deployment, which is currently done manually by Hank, would thus be described in an open-source way, possibly in a separate repo.

2. Move the backend worker containers from being self-managed Docker containers on EC2 instances to [Amazon Elastic Container Service (ECS)](https://aws.amazon.com/ecs/) + [AWS Fargate](https://aws.amazon.com/fargate/), allowing for containers to be directly managed by AWS.

3. (This one is a bit farther off.) Use an [Amazon Elastic Load Balancer (ELB)](https://aws.amazon.com/elasticloadbalancing/) of the Application Load Balancer variety in order to distribute compute jobs to different worker nodes.
    - This would potentially replace much of the functionality in [compute.py](https://github.com/OpenSourcePolicyCenter/PolicyBrain/blob/master/webapp/apps/taxbrain/compute.py), as the webapp would no longer have to keep track of a list of worker nodes and distribute the load among them. Instead, it would query a single address that would remain constant, and receive a [cookie](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-target-groups.html#sticky-sessions) allowing it to ping the worker assigned to each job, without having to worry about which worker it is.
    - This would make it easier to eventually expose the backend as an API accessible to end-users, as the load balancer can do things like rate-limiting.

I look forward to hearing feedback regarding these proposed changes, and am greatly open to suggestions.

hdoupe:
@lucassz Thanks for laying out this plan. I'm not very familiar with Terraform, but from what I've read, it looks like a pretty cool tool and will address our needs ([here](https://blog.gruntwork.io/why-we-use-terraform-and-not-chef-puppet-ansible-saltstack-or-cloudformation-7989dad2865c)'s a blog series that I found helpful for getting up to speed).

As we've discussed, ECS seems like a good option for setting up servers, deploying docker containers to them, and scaling them up and down. 

In my head, it seems like (1) and (2) are substitutes. What are the advantages of using both instead of only Terraform or only ECS?

[...]

PR #914, opened 2018-07-02
Load Redis client from existing environment variable

lucassz:
Load Redis client from existing environment variable instead of hardcoding an unchangeable hostname for it. Necessary for #913 in an auxiliary way, and seems like a good practice. It's not totally clear to me what the distinction between `CELERY_BROKER_URL` and `CELERY_RESULT_BACKEND` (which we set to the same value in practice) is, so do let me know if it should be the other one.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -25,9 +25,9 @@ for information on how to set this up on your operating system.
     2. run
     ```
     export PB_VERSION='type current PolicyBrain version here'
-    docker build -t distributed:$PB_VERSION ./
-    docker build -t celery:$PB_VERSION --file Dockerfile.celery ./
-    docker build -t flask:$PB_VERSION --file Dockerfile.flask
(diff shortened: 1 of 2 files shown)


hdoupe:
@lucassz this looks good to me. Thanks for doing the update. 

> It's not totally clear to me what the distinction between `CELERY_BROKER_URL` and `CELERY_RESULT_BACKEND` (which we set to the same value in practice) is, so do let me know if it should be the other one.

Some where in the celery docs it discusses ways to combine redis and rabbitmq or something like that. However, for us, just using redis gets the job done.

lucassz:
Added a semi-related update to `DOCKER.md`.

hdoupe:
@lucassz is #914 ready to merge?

lucassz:
@hdoupe Yes, looks good to me.

hdoupe:
Thanks for the contribution @lucassz!

PR #919, opened 2018-07-05
PEP8 fixes

lucassz:
PEP8 fixes, done using [autopep8](https://github.com/hhatto/autopep8) and a lot of manual tweaking for more intuitive indentation than the automated tool gives. I'm looking forward to discussion about the advantages and disadvantages of this. The main possible disadvantage is that it makes code history less easily navigable using Git. The main advantage is that this would bring us to an equilibrium point of maximum readability, and it would not be strenuous after that to make sure that new commits still follow the standard, so it would be a one-time adaptation.


webapp/apps/btax/tests/test_views.py
```diff
@@ -58,11 +61,15 @@ def test_btax_post(self):
...
-        self.assertTrue(response.url[:link_idx+1].endswith("ccc/"))
+        self.assertTrue(response.url[:link_idx + 1].endswith("ccc/"))
 
     def test_btax_nodes_down(self):
+<<<<<<< HEAD
         #Monkey patch to mock out running of compute jobs
         import sys
+=======
+        # Monkey patch to mock out running of compute jobs
+>>>>>>> 51b2aa3... PEP8 fixes
```
hdoupe: A merge conflict marker snuck in here.

lucassz:
@hdoupe Thanks for the catch, I fixed a few merge conflicts and amended the commit. I had to do a lot of manual merging because I was first working on top of #917, then realized it was probably better to separate the two until the other one is merged.

hdoupe:
Thanks for the contribution @lucassz

PR #926, opened 2018-08-13
In Python 3, '01' causes a syntax error when evaled via ast.literal_eval

### LINKED ISSUE #924: Submitting the TaxBrain form with parameters with an invalid type does not display an error
Submitting the input form in TaxBrain with a parameter not respecting the time, such as the string "foo" instead of a float, simply reloads that input page when submitting it, not displaying any kind of error or hint that there is one.

hdoupe:
`ast.literal_eval` throws a syntax error in Python 3 when passed a string like `"01"`. In Python 2, `ast.literal_eval` parses this string to `1`. The behavior has been changed to return the original, unparsed value if an exception is thrown. Errors such as this should be caught by the initial input validation logic. The only reason that this occurred is that bad data got through the initial form validation.

Note that there was a bug where the wrong variable was returned and a `NameError` was thrown. This has been fixed.

While working on the issue above, I noticed that the input value validation logic needed some serious work. I spent some time trying to get the `pyparser` library to work as expected, but I was unsuccessful. Thus, I wrote an `is_safe` function that tests:
- how long is the input string?
for each value in a comma separated value string:
- is this value an integer or float?
- is this value a boolean value?
- is this value a reverse character (<)?
- is this value a wildcard character (*)?

Also, this resolves #924

### DIFF
--- a/webapp/apps/taxbrain/helpers.py
+++ b/webapp/apps/taxbrain/helpers.py
@@ -50,7 +50,8 @@ def is_safe(s):
     Returns:
         success: whether value is "safe" or not
     """
-    assert len(s) < 100
+    if len(s) > 100:
+        return False
     parsers = [VALUE, BOOL, REVERSE]
     tokens = s.split(',')
     success = [False] * len(tokens)
--- a/webapp/apps/taxbrain/helpers.py
+++ b/webapp/apps/taxbrain/helpers.py
@@ -3,6 +3,7 @@ import pyparsing as pp
 import sys
 import six
 import re
+import ast
 
 # Mock some module for imports because we can't fit them on Heroku slugs
 import sys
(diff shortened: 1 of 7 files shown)


webapp/apps/taxbrain/helpers.py
```diff
@@ -30,18 +31,45 @@
...
+    - is the string an integer or float?
+    - is the string a wildcard (*)?
+    - is the string a reverse character (<)?
+    If one of the tokens does not answer one of the above questions in the
+    affirmative, then it is deemed "not safe"
+
+    Returns:
+        success: whether value is "safe" or not
+    """
+    assert len(s) < 100
```
lucassz: I think it would be better to raise some other kind of exception here and only use `assert` for internal invariants.

webapp/apps/taxbrain/helpers.py
```diff
@@ -30,18 +31,45 @@
...
+                if parser == VALUE:
+                    # make sure `ast.literal_eval` can parse
+                    # throws ValueError or SyntaxError on failure
+                    ast.literal_eval(token_strip)
+            except (pp.ParseException, ValueError, SyntaxError):
+                pass
+            else:
+                success[i] = True
+                break
+    return all(success)
```
lucassz: I like the design of this, it's very clean.

webapp/apps/taxbrain/helpers.py
hdoupe: Ah, good catch. I changed this to return `False` instead. That's what I meant to do in the first place.

webapp/apps/taxbrain/helpers.py
```diff
@@ -50,7 +50,8 @@ def is_safe(s):
     Returns:
         success: whether value is "safe" or not
     """
-    assert len(s) < 100
+    if len(s) > 100:
+        return False
```
hdoupe: @lucassz is this better? Or do you think there should be a special exception?

webapp/apps/taxbrain/helpers.py
hdoupe: Thanks. The use of `ast.literal_eval` in addition to the `VALUES` validator is a little clunky, but I'm not sure how else to catch `"01"` before it gets to `parse_values`.

hdoupe:
Thanks for the review @lucassz.

PR #931, opened 2018-08-30
Remove drip.co marketing

hdoupe:


hdoupe:
I'm going to merge #931 into the v1.7.0 branch. This is a quick fix that needs to go now so that the drip.co subscription can be canceled.

PR #936, opened 2018-10-31
Add subscribed landing page and hide registration app

hdoupe:
This PR adds a landing page that users will be redirected to once they sign up for our mailing list. The user auth components of the app were also disabled since they are no longer used.

Here is a screenshot of the landing page:

screen shot 2018-10-31 at 2 18 03 pm

cc @Peter-Metz @MattHJensen


hdoupe:
Closing #936. Need to base this on the stable branch.

PR #948, opened 2018-11-09
No longer need to scale numbers

### LINKED ISSUE #947: Zeroes reported on Static Results page
Beginning sometime yesterday, I've started getting results pages with mostly zeroes in the analysis tables.  The "total liabilities by calendar year (change)" table has nothing but zeroes except in the two right-most columns.  The difference by expanded income bin table has all zeroes in the "all tax units," "tax units with tax cut," "tax units with tax increase," and "total tax difference" columns.

hdoupe:
Resolves #947. Thanks for the bug report @siervicul.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -4,6 +4,18 @@ Go
 [here](https://github.com/OpenSourcePolicyCenter/PolicyBrain/pulls?q=is%3Apr+is%3Aclosed)
 for a complete commit history.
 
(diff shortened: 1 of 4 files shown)


hdoupe:
screen shot 2018-11-09 at 5 00 58 pm

Before the fix, all numbers were zero. (Number budget years was set to two for dev purposes)

